PR #5, opened 2020-08-10
[Bug #1356] Fix issues with resource removal caused by detached entit…

ledsoft:
…y reference not containing name required by DocumentManager.

https://kbss.felk.cvut.cz/redmine/issues/1356


blcham (approved):
seems ok

PR #11, opened 2020-10-08
1333 multilingual attributes

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1333

This is backend support for multilingual attributes and its application on Term label.


psiotwo (approved):
lgtm

PR #12, opened 2020-10-15
1333 multilingual attributes term

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1333 - use multilingual strings for Term definition, altLabels and hiddenLabels.


psiotwo (approved):
lgtm

PR #30, opened 2021-01-15
1469 terms with parents are roots bug

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1469


blcham (approved):
LGTM

although not sure whether it wouldn't be easier to infer hasTopConcept property instead of managing it (but I don't see the whole context of it)

ledsoft:
Good point, we may revisit this.

PR #31, opened 2021-01-18
[R1378] support for changing documents for vocabulary

psiotwo:


src/main/java/cz/cvut/kbss/termit/service/repository/VocabularyRepositoryService.java
```diff
@@ -75,6 +82,19 @@ protected void preUpdate(Vocabulary instance) {
         verifyVocabularyImports(instance);
     }
 
+    @Override
+    @Transactional
+    public Vocabulary update(Vocabulary instance) {
+        final Document document = instance.getDocument();
+        if (document != null) {
+            // to prevent trying to persist an unmanaged object
+            Document d = (Document) resourceService.findRequired(document.getUri());
```
ledsoft: Should be enough to use a reference (does not load all attributes).

ledsoft (dismissed):
See the comment, otherwise looks ok.

ledsoft (approved):

PR #32, opened 2021-01-20
[R1378] experiment with document context switch

psiotwo:
The goal is to be able to transfer a document from a context to another context (from voc to default when detaching the doc and from default to voc when attaching to voc).

The logic employed:
1. removing the original vocabulary.document
2. evict the cache
3. repersist the original vocabulary.document in the defaultCtx (detaching the doc, so moving to defaultCtx)
4. persist the new vocabulary.document in the vocabularyCtx (attaching the doc)

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/service/repository/ResourceRepositoryService.java
+++ b/src/main/java/cz/cvut/kbss/termit/service/repository/ResourceRepositoryService.java
@@ -181,7 +181,7 @@ public class ResourceRepositoryService extends BaseAssetRepositoryService<Resour
         final Document dOriginal = vOriginal.getDocument();
         final Document document = vNew.getDocument();
 
-        if ( dOriginal == document || ( dOriginal != null && dOriginal.equals(document) ) ) {
+        if (Objects.equals(dOriginal, document)) {
             return;
         }
 
--- a/src/main/java/cz/cvut/kbss/termit/service/repository/ResourceRepositoryService.java
+++ b/src/main/java/cz/cvut/kbss/termit/service/repository/ResourceRepositoryService.java
@@ -179,6 +179,12 @@ public class ResourceRepositoryService extends BaseAssetRepositoryService<Resour
      */
     public void rewireDocumentsOnVocabularyUpdate(final Vocabulary vOriginal, final Vocabulary vNew) {
         final Document dOriginal = vOriginal.getDocument();
+        final Document document = vNew.getDocument();
+
+        if ( dOriginal == document || dOriginal.equals(document) ) {
+            return;
+        }
+
         if ( dOriginal != null ) {
             // remove from vocabulary ctx
             resourceDao.removeDocumentFromContext(dOriginal);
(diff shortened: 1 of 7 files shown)


src/main/java/cz/cvut/kbss/termit/service/repository/ResourceRepositoryService.java
```diff
@@ -169,4 +169,31 @@ private void removeFromParentDocumentIfFile(Resource instance) {
...
     }
+
+    /**
+     * Moves the triples of the document from the original vocabulary to the default context and from the
+     * new vocabulary to the context of this vocabulary.
+     *
+     * @param vOriginal original version of the vocabulary (before update)
+     * @param vNew new version of the vocabulary (after update)
+     */
+    public void rewireDocumentsOnVocabularyUpdate(final Vocabulary vOriginal, final Vocabulary vNew) {
```
ledsoft: This seems quite inefficient to me. What if the document is the same? Then it will be moved around pointlessly.

ledsoft:
A sidenote - since documents are now much more loosely coupled with vocabularies, wouldn't it make sense to have all resources in a single separate context? These little dances with contexts would not be necessary anymore.

psiotwo:
> A sidenote - since documents are now much more loosely coupled with vocabularies, wouldn't it make sense to have all resources in a single separate context? These little dances with contexts would not be necessary anymore.

I was thinking about this option as well. It depends a bit on how this feature will be used. I still feel that most vocabularies should have a document attached to source the terms from.

src/main/java/cz/cvut/kbss/termit/service/repository/ResourceRepositoryService.java
psiotwo: Good point, thanks for noting. Will fix this case.

ledsoft:
Another thing which I don't like about the new design is that now the fact that documents are stored in vocabulary context (which is IMHO persistence detail) leaks into the business logic layer via the new DAO methods.

ledsoft (dismissed):
I have objections to the current design (see my previous comment) and would vote for reworking the resources to be stored in a dedicated context. But otherwise, this PR looks OK to me.

psiotwo:
> Another thing which I don't like about the new design is that now the fact that documents are stored in vocabulary context (which is IMHO persistence detail) leaks into the business logic layer via the new DAO methods.

I agree with this. I have partially refactored the ResourceDao to contain just one "leaking" method and tried to reformulate its contract to be less implementation dependent.

psiotwo:
> I have objections to the current design (see my previous comment) and would vote for reworking the resources to be stored in a dedicated context. But otherwise, this PR looks OK to me.

I have no strong argument for keeping documents in vocabulary context, yet my "feeling" is that loose coupling of documents and vocabularies is a bit artificial and most scenarios will expect vocabulary to be bound to a document (and yes, not vice-versa). 

Having a separate context for resources is a viable option too. 

I will merge this one as is (otherwise I never finish this bit). And I am ready to discuss these options.

PR #42, opened 2021-02-25
1503 backend authorization

psiotwo:
- updated the user role resolution - each user role has a map of roles it grants privileges from. Upon user retrieval, these granted roles are inferred for the user.
- securing controllers

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/rest/BaseController.java
+++ b/src/main/java/cz/cvut/kbss/termit/rest/BaseController.java
@@ -35,7 +35,7 @@ import static cz.cvut.kbss.termit.util.Constants.*;
  * <p>
  * Will be used to define general security for the public API.
  */
-@PreAuthorize("hasRole('" + SecurityConstants.ROLE_USER + "')")
+@PreAuthorize("hasRole('" + SecurityConstants.ROLE_RESTRICTED_USER + "')")
 public class BaseController {
 
     protected final IdentifierResolver idResolver;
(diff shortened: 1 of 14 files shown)


src/main/java/cz/cvut/kbss/termit/rest/BaseController.java
```diff
@@ -35,7 +35,7 @@
  * <p>
  * Will be used to define general security for the public API.
  */
-@PreAuthorize("hasRole('" + SecurityConstants.ROLE_USER + "')")
+@PreAuthorize("hasRole('" + SecurityConstants.ROLE_RESTRICTED_USER + "')")
```
ledsoft, line 38: I would probably rather go with `hasAnyRole` of the supported ones and not inherit the roles (it feels weird that an admin inherits restricted role, but that is probably just a naming thing).

ledsoft (approved):
See the one comment as food for thought, otherwise look perfectly fine.

src/main/java/cz/cvut/kbss/termit/rest/BaseController.java
psiotwo: I was considering this option as well, mainly because in the current version inferring the roles in the backend changes makes the semantics of the role on backend different from the semantics of the role in data.

But then I decided for this as I originally anticipated many more security annotations and wanted to avoid repeating the roles many times.

Anyway, being aware of these options, I will keep the implementation as it is.

PR #47, opened 2021-03-04
1516 public api ignore token

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1516


src/main/java/cz/cvut/kbss/termit/security/JwtAuthorizationFilter.java
```diff
@@ -76,17 +79,49 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
+                return;
+            } else {
+                unauthorizedRequest(request, response, e);
+                return;
+            }
+        } catch (DisabledException | LockedException | UsernameNotFoundException e) {
+            unauthorizedRequest(request, response, e);
             return;
         }
         chain.doFilter(request, response);
```
psiotwo: why not to put chain.doFilter into the try block and avoid the return statements at all?

psiotwo (dismissed):
lgtm, see comment

PR #49, opened 2021-03-05
1455 sorted subterms

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1455

Note that terms whose label starts with a Czech character get pushed to the end anyway due to the way SPARQL ORDER BY works. If we wanted the ordering to take locale into account, we would have to write a custom ordering function for SPARQL or sort the terms in code.


psiotwo (approved):
ok

PR #59, opened 2021-03-29
[R-#1529] fixing missing front-facing URL for annotace

psiotwo:
Hot fix for https://kbss.felk.cvut.cz/redmine/issues/1529 before https://kbss.felk.cvut.cz/redmine/issues/1497 is resolved.

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/util/ConfigParam.java
+++ b/src/main/java/cz/cvut/kbss/termit/util/ConfigParam.java
@@ -25,9 +25,9 @@ public enum ConfigParam {
     REPOSITORY_URL("repository.url"),
 
     /**
-     * URL of the main application repository.
+     * Public URL of the main application repository.
      */
-    REPOSITORY_PUBLIC_URL("repository.public-url"),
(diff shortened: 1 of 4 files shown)


src/main/java/cz/cvut/kbss/termit/util/ConfigParam.java
```diff
@@ -24,6 +24,11 @@
      */
     REPOSITORY_URL("repository.url"),
 
+    /**
+     * URL of the main application repository.
```
ledsoft: The comment is duplicate of the REPOSITORY_URL, which is especially unfortunate since it gives absolutely no explanation as to why there is another repository URL parameter.

src/main/java/cz/cvut/kbss/termit/util/ConfigParam.java
```diff
@@ -24,6 +24,11 @@
...
+     */
+    REPOSITORY_PUBLIC_URL("repository.public-url"),
```
ledsoft: Other parameters use camelCase. It would be nice to stay consistent when naming the parameters...

PR #69, opened 2021-05-01
1580 exact match new

psiotwo:
https://kbss.felk.cvut.cz/redmine/issues/1580

@ledsoft

rewritten as per your implementation
removal on inferred side not working, have seen your todo, so expecting you are about to handle this.


rulesets/rules-termit-graphdb.pie
```diff
@@ -91,11 +91,38 @@ Rules
...
+      source <skos:exactMatch> this
+      -------------------------------
+      this <skos:exactMatch> source
+
+  Id: skos_exact_match_transitivity
+
+      x <skos:exactMatch>    y           [Constraint x != y, x != z, y != z]
+      y <skos:exactMatch>    z
+      -------------------------------
+      x <skos:exactMatch>    z
```
ledsoft, line 126: Shouldn't we unify pdp_term_9, pdp_term_10 and skos_exact_match_symmetry and skos_exact_match_transitivity rules? E.g., the related-ones expect the subject/object to be a Concept while the exactMatch not (curiously, the corresponding SPIN rule does).

psiotwo: skos:Concept is the domain of these relationships. I removed the skos:Concept check from the SPIN rule.

ledsoft (approved):
LGTM

PR #72, opened 2021-05-08
1549 skos related

ledsoft:
Backend for https://kbss.felk.cvut.cz/redmine/issues/1549

Note that the application ruleset has been adjusted, so it has to be applied on the repository before verifying this PR.

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/service/term/OrphanedInverseTermRelationshipRemover.java
+++ b/src/main/java/cz/cvut/kbss/termit/service/term/OrphanedInverseTermRelationshipRemover.java
@@ -13,6 +13,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 
 @Component
 public class OrphanedInverseTermRelationshipRemover {
@@ -29,6 +30,7 @@ public class OrphanedInverseTermRelationshipRemover {
     public void removeOrphanedInverseTermRelationships(Term update, Term original) {
         removeOrphanedRelated(update, original);
         removeOrphanedRelatedMatch(update, original);
+        removeOrphanedExactMatches(update, original);
     }
 
     private Set<TermInfo> determineOrphaned(Set<TermInfo> newValue, Set<TermInfo> originalValue) {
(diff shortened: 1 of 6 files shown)


src/main/java/cz/cvut/kbss/termit/service/term/OrphanedInverseTermRelationshipRemover.java
```diff
@@ -0,0 +1,75 @@
...
+    private void removeOrphanedRelated(Term update, Term original) {
+        LOG.trace("Removing orphaned inverse related relationships of term {}.", update);
+        final TermInfo tiUpdate = new TermInfo(update);
+        final Set<TermInfo> orphaned = determineOrphaned(update.getInverseRelated(), original.getInverseRelated());
+        LOG.trace("Found {} orphaned related to remove.", orphaned);
+        orphaned.forEach(o -> {
+            final Optional<Term> tOpt = termDao.find(o.getUri());
+            assert tOpt.isPresent();
+            final Term t = tOpt.get();
+            if (t.getRelated() != null) {
```
psiotwo: When t.getRelated() is null here?

src/main/java/cz/cvut/kbss/termit/service/term/OrphanedInverseTermRelationshipRemover.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+
+@Component
+public class OrphanedInverseTermRelationshipRemover {
```
psiotwo, line 19: The same logic seems to be applicable for exactMatches (and later e.g. for closeMatches) - parameterizing the logic might be valuable.

src/main/java/cz/cvut/kbss/termit/service/term/OrphanedInverseTermRelationshipRemover.java
```diff
@@ -0,0 +1,75 @@
...
+    public void removeOrphanedInverseTermRelationships(Term update, Term original) {
+        removeOrphanedRelated(update, original);
+        removeOrphanedRelatedMatch(update, original);
+    }
+
+    private Set<TermInfo> determineOrphaned(Set<TermInfo> newValue, Set<TermInfo> originalValue) {
+        if (originalValue == null || originalValue.isEmpty()) {
+            return Collections.emptySet();
+        }
+        final Set<TermInfo> orphaned = new HashSet<>(originalValue);
```
psiotwo, line 40: looks like retainall, but you probably need this to avoid NPEs.

psiotwo (commented):
Some questions are inline.

I didn't test the functionality.

src/main/java/cz/cvut/kbss/termit/service/term/OrphanedInverseTermRelationshipRemover.java
ledsoft: I don't see it. I need a diff of the two sets - need to keep, which is AFAIK the opposite of `retainAll`. But anyway yes, I also need to avoid NPEs.

src/main/java/cz/cvut/kbss/termit/service/term/OrphanedInverseTermRelationshipRemover.java
ledsoft: This is the case of related getting inferred based on relatedMatch.

src/main/java/cz/cvut/kbss/termit/service/term/OrphanedInverseTermRelationshipRemover.java
ledsoft: Wasn't part of this task but ok, implemented.

psiotwo (approved):
lgtm - checked just code

PR #75, opened 2021-05-13
1586 1590 extend readonly term attributes

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1590
https://kbss.felk.cvut.cz/redmine/issues/1586

Did them in one go as they are closely related.


psiotwo (commented):
vocabulary attribute is still missing in the response of https://termit.fel.cvut.cz/service/ipr/termit/rest/public/vocabularies/psp-2016/terms/?namespace=http://onto.fel.cvut.cz/ontologies/slovn%C3%ADk/datov%C3%BD/

psiotwo (approved):
lgtm, didn't test

PR #76, opened 2021-05-13
1546

psiotwo:
https://kbss.felk.cvut.cz/redmine/issues/1546


ledsoft (approved):
LGTM, at least until we decide what to do with https://kbss.felk.cvut.cz/redmine/issues/1550

PR #79, opened 2021-06-08
Ehcache

ledsoft:
Introduces Ehcache to improve the performance of resource and vocabulary list retrieval.

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/service/repository/ResourceRepositoryService.java
+++ b/src/main/java/cz/cvut/kbss/termit/service/repository/ResourceRepositoryService.java
@@ -31,6 +31,8 @@ import cz.cvut.kbss.termit.util.ConfigParam;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.cache.annotation.CacheEvict;
+import org.springframework.cache.annotation.Cacheable;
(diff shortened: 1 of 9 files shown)


src/main/java/cz/cvut/kbss/termit/service/repository/ResourceRepositoryService.java
```diff
@@ -70,6 +74,30 @@ public ResourceRepositoryService(Validator validator, ResourceDao resourceDao,
         return resourceDao;
     }
 
+    @Cacheable
+    @Override
+    public List<Resource> findAll() {
+        return super.findAll();
+    }
+
+    @CacheEvict(allEntries = true)
```
psiotwo, line 83: why these methods evict all entries?

psiotwo (commented):
see comment

src/main/java/cz/cvut/kbss/termit/service/repository/ResourceRepositoryService.java
ledsoft: See javadoc to `CacheConfig` - there is actually only one entry in the cache and it contains the whole list of elements.

psiotwo: ok, makes sense.

psiotwo (approved):

PR #81, opened 2021-06-10
1542 get term count

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1542

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/persistence/dao/VocabularyDao.java
+++ b/src/main/java/cz/cvut/kbss/termit/persistence/dao/VocabularyDao.java
@@ -49,7 +49,8 @@ public class VocabularyDao extends AssetDao<Vocabulary> implements SupportsLastM
     private final ApplicationContext context;
 
     @Autowired
-    public VocabularyDao(EntityManager em, Configuration config, DescriptorFactory descriptorFactory, ChangeRecordDao changeRecordDao, ApplicationContext context) {
+    public VocabularyDao(EntityManager em, Configuration config, DescriptorFactory descriptorFactory,
+                         ChangeRecordDao changeRecordDao, ApplicationContext context) {
         super(Vocabulary.class, em, config, descriptorFactory);
         this.changeRecordDao = changeRecordDao;
         refreshLastModified();
@@ -98,10 +99,10 @@ public class VocabularyDao extends AssetDao<Vocabulary> implements SupportsLastM
         Objects.requireNonNull(entity);
         try {
             return em.createNativeQuery("SELECT DISTINCT ?imported WHERE {" +
-                    "?x ?imports+ ?imported ." +
-                    "}", URI.class)
-                    .setParameter("imports", URI.create(cz.cvut.kbss.termit.util.Vocabulary.s_p_importuje_slovnik))
-                    .setParameter("x", entity.getUri()).getResultList();
+                "?x ?imports+ ?imported ." +
+                "}", URI.class)
+                     .setParameter("imports", URI.create(cz.cvut.kbss.termit.util.Vocabulary.s_p_importuje_slovnik))
+                     .setParameter("x", entity.getUri()).getResultList();
         } catch (RuntimeException e) {
             throw new PersistenceException(e);
         }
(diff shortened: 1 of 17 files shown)


src/main/java/cz/cvut/kbss/termit/persistence/dao/VocabularyDao.java
```diff
@@ -49,7 +49,8 @@
     private final ApplicationContext context;
 
     @Autowired
-    public VocabularyDao(EntityManager em, Configuration config, DescriptorFactory descriptorFactory, ChangeRecordDao changeRecordDao, ApplicationContext context) {
+    public VocabularyDao(EntityManager em, Configuration config, DescriptorFactory descriptorFactory,
```
psiotwo, line 55: why this reformatting?

src/main/java/cz/cvut/kbss/termit/persistence/dao/VocabularyDao.java
```diff
@@ -223,16 +224,28 @@ public void refreshLastModified(RefreshLastModifiedEvent event) {
...
-                " }", URI.class).setParameter("type", URI.create(SKOS.CONCEPT))
-                .setParameter("vocabulary", vocabulary).getResultList();
+            "GRAPH ?vocabulary { " +
+            "?term a ?type ;" +
+            "}" +
+            "?term ?inVocabulary ?vocabulary ." +
+            " }", URI.class).setParameter("type", URI.create(SKOS.CONCEPT))
+                                  .setParameter("vocabulary", vocabulary).getResultList();
         return terms.stream().flatMap(tUri -> {
             final Term t = new Term();
```
psiotwo, line 239: Not a fancy pattern - the findAll method only does t.getUri anyway.

src/main/java/cz/cvut/kbss/termit/rest/VocabularyController.java
```diff
@@ -129,6 +129,17 @@ private URI resolveVocabularyUri(String fragment, String namespace) {
         return vocabularyService.getChangesOfContent(vocabulary);
     }
 
+    /**
+     * Gets the number of terms in the vocabulary with the specified identification
+     */
+    @PreAuthorize("permitAll()")
+    @GetMapping(value = "/{fragment}/terms/count")
```
psiotwo: Is this going to collide with a term with fragment 'count'? 

We have some other occurrences of this at other places of the code, but we should not proliferate this issue.

Why not to use e.g. X-Total-Count header ?

src/test/java/cz/cvut/kbss/termit/persistence/dao/VocabularyDaoTest.java
```diff
@@ -180,7 +180,7 @@ void updateGlossaryMergesGlossaryIntoCorrectRepositoryContext() {
         });
 
         final Glossary result = em.find(Glossary.class, vocabulary.getGlossary().getUri(),
-                descriptorFactory.glossaryDescriptor(vocabulary));
+            descriptorFactory.glossaryDescriptor(vocabulary));
```
psiotwo, line 183: Again, many style changes - I thus expect that the previous version was not formatted properly w.r.t .editorconfig.

psiotwo (commented):
The code seems almost OK, but please recheck the REST API design first.

src/main/java/cz/cvut/kbss/termit/rest/VocabularyController.java
ledsoft: I like the X-Total-Count header solution, there is just one problem: the tree select, which loads terms from the backend, uses the `/roots` endpoint. In that case, returning the total number of terms would IMHO be misleading and returning just the number of roots would not be sufficient for the functionality.

psiotwo: How is this ticket connected to the tree select component?  Isn't it just about some badge with the count in the tab outside of the tree select? For just count display it might be enough to have "HEAD /" endpoint with the X-Total-Count header

src/main/java/cz/cvut/kbss/termit/persistence/dao/VocabularyDao.java
ledsoft: I don't understand. Besides, this method has been there for ages, it appears changed only due to the formatting.

src/test/java/cz/cvut/kbss/termit/persistence/dao/VocabularyDaoTest.java
ledsoft: Hopefully fixed with updated .editorconfig.

src/main/java/cz/cvut/kbss/termit/persistence/dao/VocabularyDao.java
ledsoft: Hopefully fixed with updated .editorconfig.

src/main/java/cz/cvut/kbss/termit/persistence/dao/VocabularyDao.java
psiotwo: Yeah, true, just reformatting.

My point was that the key - tUri - seems to be enough to find a list of terms, no need to pass the whole asset. We use this e.g. in VocabularyDao. But we are not consistent through Termit with this.

ledsoft: I see. Yes, the Term URI is enough. However, I try to use domain objects in the APIs throughout the application (yes, probably not everywhere), so that is the reason for the pattern here.

psiotwo (approved):
lgtm

PR #82, opened 2021-06-18
1597 config refactoring

psiotwo:
https://kbss.felk.cvut.cz/redmine/issues/1597

Changes:
- defaults are provided in YAML only
- all TermIt properties are prefixed with 'termit' - build pipelines + Docker still needs to be updated.


ledsoft (dismissed):
TLTR

PR #83, opened 2021-06-21
[R-#1551] changing export type of a type from skos:broader to rdf:type; minor renaming of query variables

psiotwo:
https://kbss.felk.cvut.cz/redmine/issues/1551


ledsoft (approved):
LGTM

PR #85, opened 2021-06-30
Suppress logging of the UnsupportedOperationException

ledsoft:
Suppress logging of the UnsupportedOperationException thrown by the `TermRepositoryService.getUnusedTermsInVocabulary` as it was plaguing the logs with repeated useless stack traces.

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/exception/TermItException.java
+++ b/src/main/java/cz/cvut/kbss/termit/exception/TermItException.java
@@ -21,37 +21,18 @@ package cz.cvut.kbss.termit.exception;
  */
 public class TermItException extends RuntimeException {
 
-    protected final boolean suppressLogging;
-
     protected TermItException() {
-        this.suppressLogging = false;
     }
 
     public TermItException(String message) {
-        this(message, false);
-    }
-
-    public TermItException(String message, boolean suppressLogging) {
         super(message);
-        this.suppressLogging = suppressLogging;
     }
 
     public TermItException(String message, Throwable cause) {
         super(message, cause);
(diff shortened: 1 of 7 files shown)


src/main/java/cz/cvut/kbss/termit/exception/TermItException.java
```diff
@@ -24,18 +21,37 @@
  */
 public class TermItException extends RuntimeException {
```
psiotwo: I would solve it by having a dedicated marker subclass/interface for not logged exception. It would be less flexible, but would leave the code intact. But this is OK as well.

psiotwo: One thing - having the same exception logged at some point and not logged at another point seems a bit confusing to me. Maybe the exception class/interface would avoid this potential misuse.

src/main/java/cz/cvut/kbss/termit/service/repository/TermRepositoryService.java
```diff
@@ -320,33 +321,33 @@ public void remove(Term instance) {
 
         if (!ai.isEmpty()) {
             throw new TermRemovalException(
-                "Cannot delete the term. It is used for annotating resources : " +
-                    ai.stream().map(TermAssignments::getResourceLabel).collect(
-                        joining(",")));
+                    "Cannot delete the term. It is used for annotating resources : " +
```
psiotwo, line 325: Again, an .editorconfig issue ?

psiotwo (dismissed):
ok, see comments

psiotwo (approved):
lgtm

PR #86, opened 2021-07-12
[R-#1595] sorting

psiotwo:
https://kbss.felk.cvut.cz/redmine/issues/1595


src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
```diff
@@ -315,6 +315,34 @@ public boolean isEmpty(Vocabulary vocabulary) {
         }
     }
 
+    private String orderSentence(String lang, String var) {
```
ledsoft, line 318: I'd prefer some explanatory comment. Especially given the complex and weird looking "cs" case.

ledsoft (approved):

PR #88, opened 2021-07-20
1548 skos broad match

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1548

Note that, due to changes in the rule set, this PR requires re-creation of repositories and data migration (replacing `?child skos:broader ?parent` assertions with `?child skos:broadMatch ?parent` assertions in case of `?child` and `?parent` being from different schemes) for existing repositories.


src/main/java/cz/cvut/kbss/termit/persistence/DescriptorFactory.java
```diff
@@ -228,14 +228,14 @@ public Descriptor termDescriptor(Vocabulary vocabulary) {
      */
     public Descriptor termDescriptor(URI vocabularyUri) {
         final EntityDescriptor descriptor = assetDescriptor(vocabularyUri);
-        final EntityDescriptor parentDescriptor = new EntityDescriptor();
+        final EntityDescriptor externalParentDescriptor = new EntityDescriptor();
```
psiotwo, line 231: Why parentDescriptor is not needed any more?

psiotwo (dismissed):
ok

PR #89, opened 2021-07-21
[Upd] Update dependencies. Fix minor post-update issues.

ledsoft:
Update dependencies.

The main reason is updating to JB4JSON-LD 0.8.8 which fixes an issue with deserialization of plural multilingual attributes. The problem was that originally each value in any language got its own instance of `MultilingualString` in the collection. However, this was registered as a change by TermIt's change tracking, even though the values haven't actually changed. The updated version attempts to deserialize multilingual string values into as few `MultilingualString` elements as possible, similarly to the way JOPA does it when loading data from the repository. This fixes the change tracking issue.

See, for example, the history of "Pojem osmnáct" in the ML test vocabulary on termit-dev.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,16 +25,16 @@
     </repositories>
 
     <properties>
-        <org.springframework.boot.version>2.4.5</org.springframework.boot.version>
-        <org.springframework.version>5.3.6</org.springframework.version>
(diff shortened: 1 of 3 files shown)


psiotwo (approved):
ok

PR #90, opened 2021-07-21
[R-#1419] fix vocabularies refresh after import

psiotwo:
- fixing refresh of vocabularies after import is finished.


src/main/java/cz/cvut/kbss/termit/persistence/dao/skos/SKOSImporter.java
```diff
@@ -89,6 +89,7 @@ public SKOSImporter(Configuration config,
     public Vocabulary importVocabulary(final boolean rename,
                                        final URI vocabularyIri,
                                        final String mediaType,
+                                       final Consumer<Vocabulary> persist,
```
ledsoft, line 93: Please, update the Javadoc as well. The meaning of the `persist` parameter is not clear from the method signature (at least not to me), so an explanation of what is expected of it would be helpful

src/main/java/cz/cvut/kbss/termit/service/repository/VocabularyRepositoryService.java
```diff
@@ -39,28 +40,35 @@
...
         this.changeRecordService = changeRecordService;
-        this.importService = importService;
         this.resourceService = resourceService;
         this.config = config.getNamespace();
     }
 
+    /**
+     * This method ensures new instances of the prototype-scoped bean are returned on every call.
+     */
+    private SKOSImporter getSKOSImporter() {
```
ledsoft, line 68: Consider using `@Lookup` for injecting prototype beans. See `TermOccurrenceResolvers`

ledsoft (dismissed):
See the comments

PR #91, opened 2021-08-04
[Fix] Ensure subterms are loaded for terms included in loading root t…

ledsoft:
…erms to prevent issues with display in tree select.


psiotwo (approved):
lgtm

PR #96, opened 2021-09-06
[R-#1619] FTS simple implementation (on top of existing DAO layer)

psiotwo:
https://kbss.felk.cvut.cz/redmine/issues/1619

This is a simple implementation. First experiments shows it is efficient enough. Will revise on real data to see whether in-query implementation is necessary at this point.

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/service/business/SearchService.java
+++ b/src/main/java/cz/cvut/kbss/termit/service/business/SearchService.java
@@ -1,28 +1,33 @@
 /**
  * TermIt
  * Copyright (C) 2019 Czech Technical University in Prague
- *
+ * <p>
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- *
+ * <p>
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- *
+ * <p>
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 package cz.cvut.kbss.termit.service.business;
 
+import cz.cvut.kbss.jopa.vocabulary.SKOS;
 import cz.cvut.kbss.termit.dto.FullTextSearchResult;
 import cz.cvut.kbss.termit.persistence.dao.SearchDao;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
+import java.net.URI;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


src/main/java/cz/cvut/kbss/termit/service/business/SearchService.java
```diff
@@ -38,9 +43,18 @@ public SearchService(SearchDao searchDao) {
      * Executes full text search in assets.
      *
      * @param searchString String to search by
+     * @param vocabularies URIs of vocabularies to search in, or null, if all vocabularies shall be searched
+     * @param termsOnly whether to return only matches related to terms (not vocabularies)
      * @return Matching assets
      */
-    public List<FullTextSearchResult> fullTextSearch(String searchString) {
-        return searchDao.fullTextSearch(searchString);
+    public List<FullTextSearchResult> fullTextSearch(String searchString, Set<URI> vocabularies, boolean termsOnly) {
```
ledsoft: Not a good idea using boolean switches, especially given the quite significantly different nature of the underlying logic. Two methods explicitly specifying the implemented functionality should be used IMHO.

src/main/java/cz/cvut/kbss/termit/service/business/SearchService.java
```diff
@@ -38,9 +43,18 @@ public SearchService(SearchDao searchDao) {
...
+        Stream<FullTextSearchResult> result = searchDao.fullTextSearch(searchString).stream();
+        if (termsOnly) {
+            result = result.filter(r -> r.getTypes().contains(SKOS.CONCEPT));
+        }
+        if (vocabularies != null) {
+            result = result.filter(r -> vocabularies.contains(r.getVocabulary()));
```
ledsoft, line 63: Possible (but admittedly unlikely) NPX in case termsOnly is false, vocabularies is not null and r is a vocabulary. In that case, contains would be called with null, which for some collection may result in NPX.

ledsoft (changes requested):
See the comments.
Also, I wonder whether simple in this context means "we will probably change it in the future" (in which case I don't understand why not do it properly right now, given that it wouldn't take much extra effort).

psiotwo:
> See the comments.
> Also, I wonder whether simple in this context means "we will probably change it in the future" (in which case I don't understand why not do it properly right now, given that it wouldn't take much extra effort).

Simple means - minimal in changes. Until there is a performance problem doing small code changes is preferable from my PoV. Yes, it might happen that more efficient implementation will be needed. In that case I would probably redesign FTS completely as even the current version of the FTS query with OPTIONALs is probably suboptimal.

src/main/java/cz/cvut/kbss/termit/service/business/SearchService.java
psiotwo: Expected uses of the method:
- vocabularies = null, termsOnly = false - implementation of FTS so far
- vocabularies = null, termsOnly = true - we don't have a use-case yet.
- vocabularies != null, termsOnly = false - we don't have a use-case yet.
- vocabularies != null, termsOnly = true - implementation of FTS for IPR

So I agree, will remove the switch and create a separate method.

ledsoft (approved):
LGTM

PR #97, opened 2021-09-07
Merge the definitionally-related terms feature into the public code base

ledsoft:
This PR publishes the previously closed source definitionally related terms feature in the public open source version of TermIt.

It also contains implementation of https://kbss.felk.cvut.cz/redmine/issues/1612 backend.

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/config/AppConfig.java
+++ b/src/main/java/cz/cvut/kbss/termit/config/AppConfig.java
@@ -14,12 +14,22 @@
  */
 package cz.cvut.kbss.termit.config;
 
+import cz.cvut.kbss.termit.util.AsyncExceptionHandler;
+import org.springframework.aop.interceptor.AsyncUncaughtExceptionHandler;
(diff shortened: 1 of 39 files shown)


ledsoft:
Just for context, adding messages of the original R-1612 implementation commits:

- [Task #1612] Ensure term definition text analysis can be invoked only by editors. Also some minor refactoring and explanation that the explicit analysis endpoints are legacy.
- [Task #1612] Ensure proper async processing configuration and error handling.
- [Task #1612] Analyze definitions of all terms in vocabulary on term persist. Currently invoked synchronously, will consider asynchronous processing after testing performance.
- [Task #1612] Execute text analysis on all terms in vocabulary when a term's label changes.
- [Task #1612] Invoke the text analysis of all terms in a vocabulary asynchronously. This improves performance, but more importantly, it ensures that the transaction that caused the text analysis invocation (e.g., term update, persist) finishes before the actual text analysis is executed and Annotace has up-to-date data.
-

psiotwo (approved):
too long to check

PR #108, opened 2021-10-26
1628 export matches

ledsoft:
Backend implementation of https://kbss.felk.cvut.cz/redmine/issues/1628

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/persistence/dao/skos/SKOSExporter.java
+++ b/src/main/java/cz/cvut/kbss/termit/persistence/dao/skos/SKOSExporter.java
@@ -3,13 +3,11 @@ package cz.cvut.kbss.termit.persistence.dao.skos;
 import cz.cvut.kbss.jopa.model.EntityManager;
 import cz.cvut.kbss.termit.model.Vocabulary;
 import cz.cvut.kbss.termit.util.Utils;
+import org.eclipse.rdf4j.model.IRI;
 import org.eclipse.rdf4j.model.Model;
 import org.eclipse.rdf4j.model.ValueFactory;
 import org.eclipse.rdf4j.model.impl.LinkedHashModel;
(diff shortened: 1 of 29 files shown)


src/main/java/cz/cvut/kbss/termit/rest/TermController.java
```diff
@@ -61,6 +66,10 @@ private URI getVocabularyUri(Optional<String> namespace, String fragment) {
      * @param vocabularyIdFragment Vocabulary name
      * @param namespace            Vocabulary namespace. Optional
      * @param searchString         String to filter term labels by. Optional
+     * @param includeImported      Whether to include imported vocabularies when searching for terms. Does not apply to term export. Optional, defaults to false
```
psiotwo, line 69: The number of endpoint parameters is growing a bit. Seven is reasonable max IMO.

src/main/java/cz/cvut/kbss/termit/persistence/dao/skos/SKOSExporter.java
```diff
@@ -82,6 +97,67 @@ public void exportGlossaryTerms(Vocabulary vocabulary) {
         }
     }
 
+    /**
+     * Exports terms referenced by terms from the previously exported glossary terms ({@link
+     * #exportGlossaryTerms(Vocabulary)}) via one of the specified properties.
+     * <p>
+     * Note that this expects {@link #exportGlossaryTerms(Vocabulary)} to have been called prior to this method.
```
psiotwo: I don't get why the two methods are split in this way that requires this non-computable constraint to be present. 

Both methods are called just once and in sequence.

psiotwo (commented):
Just two design-related remarks in the text

Otherwise seems OK, but there are tons of reformatting so might have missed some bit.

psiotwo (approved):
lgtm

PR #110, opened 2021-11-02
1296 optimize batch term occurrence removal

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1296

Store term occurrences in a dedicated context per target asset and drop the context on the asset removal. To prevent the need for migrating old data, added a scheduled cleanup of orphaned term occurrences that would happen at night when  TermIt is not used.

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/model/assignment/TermOccurrence.java
+++ b/src/main/java/cz/cvut/kbss/termit/model/assignment/TermOccurrence.java
@@ -8,6 +8,11 @@ import java.net.URI;
 @OWLClass(iri = Vocabulary.s_c_vyskyt_termu)
 public abstract class TermOccurrence extends TermAssignment {
 
+    /**
+     * Suffix used to identify term occurrence contexts (named graphs) in the repository.
+     */
+    public static final String CONTEXT_SUFFIX = "occurrences";
+
     private transient Double score;
 
     public TermOccurrence() {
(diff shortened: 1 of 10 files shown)


src/main/java/cz/cvut/kbss/termit/model/assignment/TermOccurrence.java
```diff
@@ -4,10 +4,16 @@
...
 import java.net.URI;
+import java.util.Objects;
 
 @OWLClass(iri = Vocabulary.s_c_vyskyt_termu)
 public abstract class TermOccurrence extends TermAssignment {
 
+    /**
+     * Suffix used to identify term occurrence contexts (named graphs) in the repository.
+     */
+    public static final String CONTEXT_SUFFIX = "occurrences";
```
psiotwo, line 15: Why not to put it to Constants?

src/main/java/cz/cvut/kbss/termit/service/repository/TermAssignmentRepositoryService.java
```diff
@@ -249,4 +250,16 @@ public TermOccurrence getRequiredReference(URI id) {
             NotFoundException.create(TermOccurrence.class.getSimpleName(), id)
         );
     }
+
+    /**
+     * Cleans up possibly orphaned term occurrences.
+     *
+     * Such occurrences reference targets whose sources no longer exist in the repository.
+     */
+    @Scheduled(cron = "0 1 1 * * ?")
```
psiotwo: Why not to put the CRON pattern to Constants?

psiotwo (dismissed):
lgtm

src/main/java/cz/cvut/kbss/termit/model/assignment/TermOccurrence.java
ledsoft: I don't see a reason for putting this highly specific value into global application constants.

psiotwo (approved):
lgtm

PR #123, opened 2022-01-31
[#183] Replace UFO type labels.

ledsoft:
Changed content types based on https://github.com/kbss-cvut/termit-ui/issues/183

### DIFF
--- a/src/main/resources/languages/language.ttl
+++ b/src/main/resources/languages/language.ttl
@@ -38,11 +38,11 @@ ufo:event-type a a-popis-dat:term ;
 
 ufo:intrinsic-trope a a-popis-dat:term ;
     rdfs:label "Vlastnost"@cs,
-        "Intrinsic Trope"@en .
+        "Aspect"@en .
 
 ufo:intrinsic-trope-type a a-popis-dat:term ;
     rdfs:label "Typ vlastnosti"@cs,
-        "Intrinsic Trope Type"@en .
+        "Aspect Type"@en .
 
 ufo:object a a-popis-dat:term ;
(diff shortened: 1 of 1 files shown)


src/main/resources/languages/language.ttl
```diff
@@ -55,8 +55,8 @@ ufo:object-type a a-popis-dat:term ;
 
 ufo:relator a a-popis-dat:term ;
     rdfs:label "Vztah"@cs,
-        "Relator"@en .
+        "Relation"@en .
 
 ufo:relator-type a a-popis-dat:term ;
     rdfs:label "Typ vztahu"@cs,
-        "Relator Type"@en .
+        "Relation Type"@en .
```
psiotwo: Relator

psiotwo (dismissed):
see comment, otherwise ok

src/main/resources/languages/language.ttl
ledsoft: So it's "Relation" but "Relator Type"?

psiotwo: No, I meant Relator, not 'Relator type'. 

But to be sure I looked again in [doc](https://dev.nemo.inf.ufes.br/seon/UFO.html) and I was wrong, it is vice versa, so:

Relation Type -> Relator
Relator Type -> Relation

ledsoft: I still don't get it. The language (as can be seen in the file) contains only "Relator" and "Relator type". So how do we change their translation?

psiotwo: see my commit

PR #127, opened 2022-03-07
1652 remove term assignment feature

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1652


src/main/java/cz/cvut/kbss/termit/dto/occurrence/ResourceTermOccurrences.java
```diff
@@ -45,7 +43,13 @@
                 @VariableResult(name = "suggested", type = Boolean.class)
         }
 ))
-public class ResourceTermOccurrences extends ResourceTermAssignments {
+public class ResourceTermOccurrences extends AbstractTermOccurrenceInfo {
```
psiotwo: OK. Yet, such changes will make any future continuation on the closed-source TermIt even less mergeable with the open-source. However, probably not a big deal anyway.

src/main/java/cz/cvut/kbss/termit/service/repository/TermOccurrenceRepositoryService.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    public TermOccurrence getRequiredReference(URI id) {
+        return termOccurrenceDao.getReference(id).orElseThrow(() ->
+                NotFoundException.create(TermOccurrence.class.getSimpleName(), id)
+        );
+    }
+
+    @Transactional
+    @Override
+    public void persistOccurrence(TermOccurrence occurrence) {
```
psiotwo: Why not align service names with other services.

psiotwo (dismissed):
lgtm

src/main/java/cz/cvut/kbss/termit/dto/occurrence/ResourceTermOccurrences.java
ledsoft: So you think the package/class renaming should be reverted?

psiotwo: Yes, I think keeping the two versions as close as possible until there is a good reason not to do it (e.g. breaking change in the code) is a better option now.

ledsoft: Ok, good point. Will do.

src/main/java/cz/cvut/kbss/termit/dto/assignment/AbstractAssignmentInfo.java
```diff
@@ -26,7 +26,7 @@
 import java.util.Objects;
 import java.util.Set;
 
-abstract class AbstractAssignmentsInfo implements HasTypes {
+abstract class AbstractAssignmentInfo implements HasTypes {
```
psiotwo: it was AbstractAssignment**s**Info

PR #131, opened 2022-04-05
Enhancement/248 optimize all terms retrieval

ledsoft:
https://github.com/kbss-cvut/termit-ui/issues/248


src/main/java/cz/cvut/kbss/termit/rest/TermController.java
```diff
@@ -85,31 +89,35 @@ private URI getVocabularyUri(Optional<String> namespace, String fragment) {
                                     @RequestParam(name = "withReferences", required = false) boolean withReferences,
                                     @RequestParam(name = "property", required = false,
                                                   defaultValue = "[]") Set<String> properties,
-                                    @RequestHeader(value = HttpHeaders.ACCEPT, required = false) String acceptType) {
+                                    @RequestHeader(value = HttpHeaders.ACCEPT, required = false,
+                                                   defaultValue = MediaType.ALL_VALUE) String acceptType) {
```
blcham, line 93: not sure why u need ALL_VALUE

blcham (approved):
LGTM

TLTR, hope I get better at this. There is one added value of having a code style with short lines. It is reviewable on a notebook when using review mode in GitHub :)

src/main/java/cz/cvut/kbss/termit/rest/TermController.java
ledsoft: It's a default, because without default value `acceptType` is null, which cause NPX.

PR #152, opened 2022-05-18
Bug/214 glossary iri consistent with vocabulary

ledsoft:
Fixes kbss-cvut/termit-ui#214.


ledsoft:
The import API has been split into two methods:
1. Import a new vocabulary. Supports IRI renaming to prevent conflicts.
2. Reimport of an existing vocabulary. Replaces content.

The original importer API was a combination of both with unclear intentions and buggy logic. The new implementation works for both cases. Frontend updates were also necessary, but they are just minor.

ledsoft:
@psiotwo has not responded in two weeks, merging.

Issue #165, opened 2022-08-01
Integration with Keycloak


ledsoft:
In order to facilitate compatibility with the [SGoV assembly line](opendata-mvcr/sgov-assembly-line), TermIt has to be able to use Keycloak as an authorization service.

However, to retain backwards compatibility, it also has to be able to run without it, using its internal authentication mechanisms for secure access to the application.

Note that this issue involves backend as well as frontend of TermIt.

PR #173, opened 2022-08-25
[#165] Keycloak support added

filip-kopecky:
Added keycloak

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,6 +41,18 @@
         <packaging>jar</packaging>
     </properties>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.keycloak.bom</groupId>
+                <artifactId>keycloak-adapter-bom</artifactId>
+                <version>19.0.1</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
     <dependencies>
         <dependency>
             <groupId>com.github.sgov</groupId>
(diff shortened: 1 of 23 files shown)


filip-kopecky:
Added profiles to the relevant test classes + removed isAuthenticated from `/configuration` endpoint. The previously failed tests should be working now. @ledsoft

pom.xml
```diff
@@ -454,6 +482,8 @@
                             <version>${org.hibernate.validator.version}</version>
                         </path>
                     </annotationProcessorPaths>
+                    <source>11</source>
+                    <target>11</target>
```
ledsoft: Replace with `java.version` from properties.

src/main/java/cz/cvut/kbss/termit/config/SecurityKeycloakConfig.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Autowired
+    public void configureGlobal(
+            AuthenticationManagerBuilder auth) throws Exception {
+
+        KeycloakAuthenticationProvider keycloakAuthenticationProvider
+                = keycloakAuthenticationProvider();
+        keycloakAuthenticationProvider.setGrantedAuthoritiesMapper(
+                new SimpleAuthorityMapper());
```
ledsoft: Replace with a custom authority mapper looking like this:
```java
public class HierarchicalRoleBasedAuthorityMapper implements GrantedAuthoritiesMapper {

    @Override
    public Collection<? extends GrantedAuthority> mapAuthorities(Collection<? extends GrantedAuthority> authorities) {
        return authorities.stream().map(a -> {
                              final UserRole role = UserRole.fromRoleName(a.getAuthority());
                              return role.getGranted();
                          }).flatMap(roles -> roles.stream().map(r -> new SimpleGrantedAuthority(r.getName())))
                          .collect(Collectors.toSet());
    }
}
```

ledsoft (changes requested):
I figured out the problem with `PreAuthorize` - TermIt internal roles work hierarchically, so `ROLE_FULL_USER` automatically includes `ROLE_RESTRICTED_USER`. Keycloak-based authorization does not work like this.
To fix it, create a custom authority mapper and pass it to `KeycloakAuthenticationProvider`. See the inline comment. After that, all the extra `PreAuthorize("isAuthenticated()")` can be removed and the PR will me much more compact.

ledsoft (approved):
LGTM

PR #191, opened 2022-10-26
Termit UI#340 skos export rdfxml

ledsoft:
Backend implementation of kbss-cvut/termit-ui#340.


blcham:
yak, why such ugly  format like rdf/xml and not turtle ...

ledsoft:
SKOS export to TTL has been in TermIt basically from the beginning:) This extension for a different serialization was requested by CAS.

PR #198, opened 2022-11-19
[kbss-cvut/termit-ui#128] Renaming of document and its files

filip-kopecky:
Added an endpoint which allows to change the name of the files. 
I wasn't able to do it via `updateResource()`, because some issue with syncing the file happened. 
The changed label wasn't properly propagated and then the file content wasn't in the right spot.

It was needed to create service function which takes the old file object and new file object and merges them together. 
I couldn't save the modified object file which comes from the `RequestBody` because it doesn't contain the `document` property. 
The reason why it is not in the object is simple. The `toJsonLd()` on the front-end side wasn't working with the `owner` property and also I was struggling with the fact that the `document` value is inferred. 

@ledsoft If you don't like this solution, I can try to make it work in a way which would only accept the object from the `RequestBody` but I think this works well.


filip-kopecky:
It turned out that no changes are needed for the desired functionality mentioned in the issue. Closing this PR

PR #204, opened 2022-12-15
Change of Change Tracking Contexts for KODI

mighantos:
I changed where change tracking data are stored for Vocabulary Contexts (VCs) so that it stores changes made in VC to a different named graph (prepared on VC creation by SGOV) than canonical changes. 

In my quick testing, the implementation did not change the behaviour of the standalone TermIt.

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/persistence/dao/changetracking/ChangeRecordDao.java
+++ b/src/main/java/cz/cvut/kbss/termit/persistence/dao/changetracking/ChangeRecordDao.java
@@ -23,9 +23,13 @@ public class ChangeRecordDao {
 
     private final EntityManager em;
 
-    public ChangeRecordDao(ChangeTrackingContextResolver contextResolver, EntityManager em) {
+    private final ChangeTrackingContextResolver changeTrackingContextResolver;
+
+    public ChangeRecordDao(ChangeTrackingContextResolver contextResolver, EntityManager em,
+                           ChangeTrackingContextResolver changeTrackingContextResolver) {
         this.contextResolver = contextResolver;
         this.em = em;
+        this.changeTrackingContextResolver = changeTrackingContextResolver;
     }
 
     /**
@@ -58,13 +62,16 @@ public class ChangeRecordDao {
         Objects.requireNonNull(asset);
         try {
             final Descriptor descriptor = new EntityDescriptor();
+            URI changeTrackingContextUri = changeTrackingContextResolver.resolveChangeTrackingContext(asset);
             descriptor.setLanguage(null);
             return em.createNativeQuery("SELECT ?r WHERE {" +
+                             "GRAPH ?ctc { ?r ?relatesTo ?asset . }" +
                              "?r a ?changeRecord ;" +
                              "?relatesTo ?asset ;" +
                              "?hasTime ?timestamp ." +
                              "OPTIONAL { ?r ?hasChangedAttribute ?attribute . }" +
                              "} ORDER BY DESC(?timestamp) ?attribute", AbstractChangeRecord.class)
+                     .setParameter("ctc", changeTrackingContextUri)
                      .setParameter("changeRecord", URI.create(Vocabulary.s_c_zmena))
                      .setParameter("relatesTo", URI.create(Vocabulary.s_p_ma_zmenenou_entitu))
                      .setParameter("hasChangedAttribute", URI.create(Vocabulary.s_p_ma_zmeneny_atribut))
--- a/src/main/java/cz/cvut/kbss/termit/persistence/dao/changetracking/ChangeRecordDao.java
+++ b/src/main/java/cz/cvut/kbss/termit/persistence/dao/changetracking/ChangeRecordDao.java
@@ -64,12 +64,11 @@ public class ChangeRecordDao {
             final Descriptor descriptor = new EntityDescriptor();
             URI changeTrackingContextUri = changeTrackingContextResolver.resolveChangeTrackingContext(asset);
             descriptor.setLanguage(null);
-            return em.createNativeQuery("SELECT ?r WHERE {" +
-                             "GRAPH ?ctc { ?r ?relatesTo ?asset . }" +
-                             "?r a ?changeRecord ;" +
-                             "?relatesTo ?asset ;" +
-                             "?hasTime ?timestamp ." +
-                             "OPTIONAL { ?r ?hasChangedAttribute ?attribute . }" +
+            return em.createNativeQuery("SELECT ?r WHERE { " +
+                             "GRAPH ?ctc { ?r ?relatesTo ?asset ; " +
+                             "?hasTime ?timestamp . " +
+                             "OPTIONAL { ?r ?hasChangedAttribute ?attribute . } } " +
+                             "?r a ?changeRecord . " +
                              "} ORDER BY DESC(?timestamp) ?attribute", AbstractChangeRecord.class)
                      .setParameter("ctc", changeTrackingContextUri)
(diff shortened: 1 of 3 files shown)


src/main/java/cz/cvut/kbss/termit/persistence/dao/changetracking/ChangeTrackingContextResolver.java
```diff
@@ -54,8 +65,19 @@ private URI resolveTermVocabulary(Term term) {
...
+
+    private URI resolveExistingChangeTrackingContext(URI vocabularyURI) {
+        Optional<URI> vocabularyContext = vocabularies.getVocabularyContext(vocabularyURI);
+        return vocabularyContext.map(uri -> em.createNativeQuery(
+                                                      "SELECT DISTINCT ?ctc WHERE { GRAPH ?vc { ?vc ?hasChangeTrackingContext ?ctc } }", URI.class)
+                                              .setParameter("vc", uri)
+                                              .setParameter("hasChangeTrackingContext",
+                                                      URI.create(
+                                                              cz.cvut.kbss.termit.util.Vocabulary.s_i_ma_kontext_sledovani_zmen))
+                                              .getSingleResult()).orElse(null);
```
ledsoft: This will throw `NoResultException` if no result is found for the query. I don't see it  being handled anywhere.

ledsoft: Also, I do not understand why you return `null` from an `Optional` and then do a null-check in the calling method. I would rather return the `Optional` and do an `orElseGet` or something like that.

src/main/java/cz/cvut/kbss/termit/persistence/dao/changetracking/ChangeTrackingContextResolver.java
```diff
@@ -39,8 +44,14 @@ public ChangeTrackingContextResolver(EntityManager em, Configuration config) {
     public URI resolveChangeTrackingContext(Asset<?> changedAsset) {
         Objects.requireNonNull(changedAsset);
         if (changedAsset instanceof Vocabulary) {
+            URI changeTrackingContextURI = resolveExistingChangeTrackingContext(changedAsset.getUri());
+            if(Objects.nonNull(changeTrackingContextURI))
+                return changeTrackingContextURI;
             return URI.create(changedAsset.getUri().toString().concat(contextExtension));
         } else if (changedAsset instanceof Term) {
+            URI changeTrackingContextURI = resolveExistingChangeTrackingContext(resolveTermVocabulary((Term) changedAsset));
```
ledsoft: I suggest adding a method that would resolve change tracking context for the term in one query instead of first getting a vocabulary and then getting its change tracking context in a separate query.

src/main/java/cz/cvut/kbss/termit/persistence/dao/changetracking/ChangeTrackingContextResolver.java
```diff
@@ -54,8 +65,19 @@ private URI resolveTermVocabulary(Term term) {
...
     }
+
+    private URI resolveExistingChangeTrackingContext(URI vocabularyURI) {
+        Optional<URI> vocabularyContext = vocabularies.getVocabularyContext(vocabularyURI);
+        return vocabularyContext.map(uri -> em.createNativeQuery(
+                                                      "SELECT DISTINCT ?ctc WHERE { GRAPH ?vc { ?vc ?hasChangeTrackingContext ?ctc } }", URI.class)
+                                              .setParameter("vc", uri)
+                                              .setParameter("hasChangeTrackingContext",
+                                                      URI.create(
+                                                              cz.cvut.kbss.termit.util.Vocabulary.s_i_ma_kontext_sledovani_zmen))
```
ledsoft: Just a minor nitpick - I prefer using the "property" versions of Vocabulary constants over the "individual" versions. So I would replace this with `cz.cvut.kbss.termit.util.Vocabulary.s_p_ma_kontext_sledovani_zmen`.

src/main/java/cz/cvut/kbss/termit/persistence/dao/changetracking/ChangeRecordDao.java
```diff
@@ -58,13 +62,16 @@ public List<AbstractChangeRecord> findAll(Asset<?> asset) {
         Objects.requireNonNull(asset);
         try {
             final Descriptor descriptor = new EntityDescriptor();
+            URI changeTrackingContextUri = changeTrackingContextResolver.resolveChangeTrackingContext(asset);
             descriptor.setLanguage(null);
             return em.createNativeQuery("SELECT ?r WHERE {" +
+                             "GRAPH ?ctc { ?r ?relatesTo ?asset . }" +
                              "?r a ?changeRecord ;" +
                              "?relatesTo ?asset ;" +
```
ledsoft: This can (and should) be removed from the query since it is already in the `GRAPH` part. Or, you can wrap the whole query in `GRAPH`. It may perform better.
```java
"SELECT ?r WHERE {" +
"GRAPH ?ctc {" +
"a ?changeRecord ;" +
"?relatesTo ?asset ;" +
"?hasTime ?timestamp ." +
"OPTIONAL { ?r ?hasChangedAttribute ?attribute . } " +
} } ORDER BY DESC(?timestamp) ?attribute"
```

ledsoft (changes requested):
See the inline comments.

src/main/java/cz/cvut/kbss/termit/persistence/dao/changetracking/ChangeRecordDao.java
mighantos: I tried wrapping the whole query in named GRAPH, but then it did not resolve that `vytvoření-entity` and `úprava-entity` are subclasses of `změna`, which the query searches for. But most certainly, I will optimize it.

src/main/java/cz/cvut/kbss/termit/persistence/dao/changetracking/ChangeTrackingContextResolver.java
mighantos: Yes, this should throw a custom exception.

mighantos: That is my bad with the `null`. It was much more complicated and I did not notice that when I simplified it. Will change to `Optional`.

src/main/java/cz/cvut/kbss/termit/persistence/dao/changetracking/ChangeTrackingContextResolver.java
mighantos: I don't think that is a good idea as it is a 2 step process with different outcomes. First, find the vocabulary and then resolve if it is in edited VC (if not, the `[vocabularyURI]/zmeny` is used). Second, If it is in an edited VC find the reference to ChangeTrackingContext (if not, throw an exception). 
If I would combine it into one query and it would not have a reference to ChangeTrackingContext or was not in VC the query would fail. Then I would still need to ask the DB what is the vocabulary and check if it is in the edited VCs. It would just transfer the double query problem from Assembly Line deployment to stand-alone deployment.

ledsoft (approved):
LGTM

src/main/java/cz/cvut/kbss/termit/persistence/dao/changetracking/ChangeTrackingContextResolver.java
ledsoft: Yeah, you are right about the types, I did not realize that.

PR #208, opened 2023-01-05
Test update for change tracking context

mighantos:
Just realized I pushed test changes to a different branch in KODI, so they were not included in the last PR #204 

The issue with the tests was that the changeRecords were persisted to the default graph and not to changeTrackingContext graph of the asset (as is the case in deployed usage).


ledsoft (approved):
LGTM

PR #219, opened 2023-03-21
Fix term deletion from all vocabulary drafts

mighantos:
Fixes deletion of a term from all vocabulary drafts and canonical version instead of just deleting from the selected draft.
Also now allows to create a term that is present in different draft version.
Term count correction to only count in the selected draft.

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
+++ b/src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
@@ -718,30 +718,6 @@ public class TermDao extends BaseAssetDao<Term> implements SnapshotProvider<Term
         }
     }
 
-    @Override
-    public boolean exists(URI id) {
-        Objects.requireNonNull(id);
-        try {
-            URI vocabularyContext = null;
-            boolean notInAnyContext = false;
-            try {
-                vocabularyContext = contextMapper.getVocabularyContext(resolveTermVocabulary(id));
-            } catch (NoResultException e) {
-                notInAnyContext = true;
-            }
-            return em.createNativeQuery("ASK { " +
-                                        (notInAnyContext ? "" : "GRAPH <"+vocabularyContext+"> { ") +
-                                        "?x a ?type . " +
-                                        (notInAnyContext ? "" : "} ") +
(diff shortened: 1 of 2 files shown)


src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
```diff
@@ -699,6 +700,30 @@ public List<TermDto> findAllIncludingImported(String searchString, Vocabulary vo
...
+    public boolean exists(URI id) {
+        Objects.requireNonNull(id);
+        try {
+            URI vocabularyContext = null;
+            boolean notInAnyContext = false;
+            try {
+                vocabularyContext = contextMapper.getVocabularyContext(resolveTermVocabulary(id));
+            } catch (NoResultException e) {
+                notInAnyContext = true;
+            }
```
ledsoft: I do not like catching the `NoResultException` here. Given the circumstances, I think it should be moved to `resolveTermContext` and when it is caught, a `NotFoundException` should be thrown, because it means the term does not exist in any vocabulary context.
Based on the current data storage architecture, it should never happen that a term is not in a vocabulary context.

ledsoft: This change would also greatly simplify the rest of this method.

ledsoft (changes requested):
See the inline comment. Otherwise it looks good to me.

src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
mighantos: I totally agree with that. I even had it done like that in the first place. But when running tests in `HtmlTermOccurrenceResolverTest` there are tests with data `rdfa-simple*.html` that don't follow that every term is in context. I tried to change the data to include the terms there in a context with `about `property. But was unsuccessful in even listing contexts present in the data with entityManager.

ledsoft: Ok, I will update the tests first and let you know.

ledsoft (approved):
LGTM

PR #225, opened 2023-03-23
Fix error with text analysis

mighantos:
Consider two drafts `A` and `B`, from the same canonical vocabulary. When rewriting definition of a term `a1` in vocabulary draft `A` and using word that is the same as term in vocabulary draft `B` that is not present in `A`. Update of term `a1` fails on text analysis.


ledsoft (approved):
LGTM

PR #229, opened 2023-04-21
Export now respects contexts

mighantos:
Export to Turtle now respects the context of the draft of selected vocabulary.


ledsoft (approved):
LGTM

PR #277, opened 2024-07-23
[Enhancement kbss-cvut/termit-ui#482] Support sending invitation links to new users

lukaskabc:
Implements kbss-cvut/termit-ui#482

- PasswordChangeService merged with UserService to avoid dependency loop  
- AdminBasedRegistrationController enabled for all profiles, it is now specific for any user registration by admin


src/main/java/cz/cvut/kbss/termit/rest/AdminBasedRegistrationController.java
```diff
@@ -56,11 +60,17 @@ public AdminBasedRegistrationController(UserService userService) {
...
 
+    @Operation(security = {@SecurityRequirement(name="bearer-key")},
+               description = "Creates a new user account. If the password is blank, the account is locked, and an email will be sent to the new user with a link to create a password.")
+    @ApiResponses({
+            @ApiResponse(responseCode = "201", description = "User created"),
+            @ApiResponse(responseCode = "409", description = "User data are invalid")
+    })
     @PreAuthorize("hasRole('" + SecurityConstants.ROLE_ADMIN + "')")
-    @PostMapping(consumes = {MediaType.APPLICATION_JSON_VALUE, JsonLd.MEDIA_TYPE})
+    @PutMapping(consumes = {MediaType.APPLICATION_JSON_VALUE, JsonLd.MEDIA_TYPE})
```
ledsoft: Why did you switch from POST to PUT for creating an account?

src/main/java/cz/cvut/kbss/termit/service/business/UserService.java
```diff
@@ -215,7 +258,7 @@ public void unlock(UserAccount account, String newPassword) {
     }
 
     private void ensureNotOwnAccount(UserAccount account, String operation) {
-        if (securityUtils.getCurrentUser().equals(account)) {
+        if (securityUtils.isAuthenticated() && securityUtils.getCurrentUser().equals(account)) {
```
ledsoft, line 260: Why did you add the first part of the check? The second part should be enough for the validation.

src/main/java/cz/cvut/kbss/termit/service/business/UserService.java
```diff
@@ -307,4 +350,47 @@ public List<RdfsResource> getManagedAssets(@NonNull UserAccount user) {
...
+    @Transactional
+    public void requestPasswordReset(String username) {
+        final UserAccount account = repositoryService.findByUsername(username)
+                                               .orElseThrow(() -> NotFoundException.create(UserAccount.class, username));
+        PasswordChangeRequest request = createPasswordChangeRequest(account);
+        passwordChangeNotifier.sendPasswordResetEmail(request);
+    }
+
+    private boolean isValid(PasswordChangeRequest request) {
+        return request.getCreatedAt().plus(securityConfig.getPasswordChangeRequestValidity()).isAfter(Instant.now());
```
ledsoft: I suggest using `Utils.timestamp` which truncates the `Instant.now` to millis for easier processing (it would be good to use it for creating PasswordChangeRequests as well).

ledsoft (commented):
See the inline comments

src/main/java/cz/cvut/kbss/termit/rest/AdminBasedRegistrationController.java
lukaskabc: Currently, `AdminBasedRegistrationController` is mapped to the same path as [`FreeRegistrationController`](https://github.com/kbss-cvut/termit/blob/da05402700f6ad7e7937e2bd46cd12657fe8c180/src/main/java/cz/cvut/kbss/termit/rest/FreeRegistrationController.java#L50) and `FreeRegistrationController.createUser` shared the same mapping with `AdminBasedRegistrationController.createUser`, wasn't a problem because only one was always active (based on profile `admin-registration-only`)  
But since I set `AdminBasedRegistrationController` as always active, I had to remove the mapping conflict.  
Another option is to change the mapping path to something like `/admin/users`.

src/main/java/cz/cvut/kbss/termit/service/business/UserService.java
lukaskabc: So that I can reuse the logic to unlock the user account when resetting the password.
There is no authenticated user at that time.

src/main/java/cz/cvut/kbss/termit/rest/AdminBasedRegistrationController.java
ledsoft: Ok, in that case I vote to change the path mapping to `/admin/users`

src/main/java/cz/cvut/kbss/termit/service/business/UserService.java
ledsoft: I don't understand. `securityUtils.getCurrentUser()` will not return user if they are not authenticated anyway, so the `securityUtils.isAuthenticated` seems unnecessary.

ledsoft: Or is it to prevent an exception from being thrown?

lukaskabc: > Or is it to prevent an exception from being thrown?

NullPointerException, yes  
or I could change the equation to `account.equals(securityUtils.getCurrentUser())`

ledsoft: Would that help, though? If you look at the `SecurityUtils.getCurrentUser` code, the NPX would probably be thrown there (or maybe ClassCastException). Anyway, I now see the reasoning. Let's keep this as is.

ledsoft (approved):
LGTM

PR #281, opened 2024-07-31
[Enhancement kbss-cvut/termit-ui#479] Allow viewing terms without label in primary language

### LINKED ISSUE #479: Allow viewing terms without label in primary language
# Description
If I import a vocabulary that has terms without labels in the instance's configured language, the terms are not visible in TermIt UI, so I cannot even fix them. I want to be able to at least view the terms, so that I can fix the missing labels.

# Use
Consider the recent CAS scenario. The original instance was running with English as primary language. We then migrated to a new server and imported the data there. This new instance is set to use Czech as primary language. However, some of the original vocabularies have terms without Czech labels. This means such terms are not visible in the new instance, even though they are in the repository.
I should be able to view the terms so that I can add the missing Czech labels directly in TermIt.

# Usefulness
1. Impossible without this feature
****

lukaskabc:
resolves kbss-cvut/termit-ui#479


ledsoft (approved):
LGTM

PR #290, opened 2024-09-03
WebSockets

lukaskabc:
Implementation of WebSockets on HTTP(S) endpoint `/termit/ws` using [STOMP protocol](https://docs.spring.io/spring-framework/reference/web/websocket/stomp/overview.html).

WebSockets will be used in future PR addressing application performance ([dev branch](https://github.com/lukaskabc/termit/tree/lukaskabc/Performance-285-throttle)).

## Authentication
WebSocket API in browsers does not allow setting custom headers for WS handshake (HTTP request for protocol upgrade), 
which leaves us with three options for how to implement authentication using the JWT token:
1. save the token as a cookie, which will be sent by the browser automatically (hopefully)
2. send the token as a query parameter
3. send it with a STOMP CONNECT message

Currently implemented as option 3, which is most reliable.  
This means any WebSocket connection is accepted, and the client has limited time (15s) to send a STOMP CONNECT message with a valid JWT token. Otherwise, the connection is closed.

1. might face issues with cookies access and browser compatibility, but it should be possible to implement for better security of the ws endpoint in the future
2. JWT would be part of history and logs, which is unacceptable

## Changes

- removed REST endpoint `/vocabularies/{localName}/validate` replaced with matching ws endpoint 
- disabled test `findTermOccurrencesSetsFoundOccurrencesAsApprovedWhenCorrespondingExistingOccurrenceWasApproved`, which is failing on Windows, until the issue is resolved

## Tests
Prepared two test runners for WebSocket testing.  
- `BaseWebSocketIntegrationTestRunner`, which starts the whole application context with an active web socket endpoint and establishes an actual web socket connection to the application. Currently used for WebSocket security test.
- `BaseWebSocketControllerTestRunner` is used to test WebSocket controllers using mocked API.

## Required changes (TODO)
- It is required to adjust proxy configuration to enable WebSocket connections support ([SockJS](https://docs.spring.io/spring-framework/reference/web/websocket/fallback.html) fallback might be implemented in the app if needed)


src/main/java/cz/cvut/kbss/termit/websocket/ResultWithHeaders.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Wrapper carrying a result from WebSocket controller
+ * including the {@link #payload}, {@link #destination} and {@link #headers} for the resulting message.
+ * <p>
+ * Do not combine with other method-return-value handlers (like {@link SendTo @SendTo})
```
ledsoft, line 18: Why? Please add a short explanation.

src/main/java/cz/cvut/kbss/termit/websocket/handler/WebSocketExceptionHandler.java
```diff
@@ -0,0 +1,213 @@
...
+import org.springframework.messaging.handler.annotation.MessageExceptionHandler;
+import org.springframework.messaging.simp.annotation.SendToUser;
+import org.springframework.messaging.simp.stomp.StompHeaderAccessor;
+import org.springframework.security.core.userdetails.UsernameNotFoundException;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.multipart.MaxUploadSizeExceededException;
+
+@SendToUser
+@ControllerAdvice
+public class WebSocketExceptionHandler {
```
ledsoft, line 42: I would prefer this to be merged with the `RestExceptionHandler` class, so that when a new exception is added, we don't have to add handlers for it in two places.

src/main/java/cz/cvut/kbss/termit/config/WebSocketConfig.java
```diff
@@ -0,0 +1,135 @@
...
+                           @Lazy SimpMessagingTemplate simpMessagingTemplate) {
+        this.configuration = configuration;
+        this.context = context;
+        this.messageAuthorizationManager = messageAuthorizationManager;
+        this.jwtAuthorizationInterceptor = jwtAuthorizationInterceptor;
+        this.jsonLdMapper = jsonLdMapper;
+        this.simpMessagingTemplate = simpMessagingTemplate;
+    }
+
+    /* WebSocket security setup (replaces @EnableWebSocketSecurity) */
```
ledsoft: Dangling comment?

src/main/java/cz/cvut/kbss/termit/config/WebAppConfig.java
```diff
@@ -153,27 +153,27 @@ public SimpleUrlHandlerMapping sparqlQueryControllerMapping() throws Exception {
     }
 
     @Bean
-    public HttpMessageConverter<?> stringMessageConverter() {
+    public HttpMessageConverter<?> termitStringHttpMessageConverter() {
```
ledsoft, line 156: Is it necessary to rename these methods? I see there are similar-named factory methods in `WebSocketConfig`, but they do not create `HttpMessageConverter`s and don't have `http` in the name. For me, this change seems unnecessary.

ledsoft (changes requested):
A couple of minor questions/suggestions

src/main/java/cz/cvut/kbss/termit/websocket/ResultWithHeaders.java
lukaskabc: I made this wrapper to be able to reply with custom headers, which would otherwise require constructing a stomp message in the controller.
The returned object is then handled by `WebSocketMessageWithHeadersValueHandler`.
Every value returned from a controller method can be handled only by a single return value handler.
Annotations like `@SendTo`/`@SendToUser` are handled by separate return value handlers, so only one can be used simultaneously.

src/main/java/cz/cvut/kbss/termit/config/WebSocketConfig.java
lukaskabc: *Section* for the methods below, should I comment each method instead?

src/main/java/cz/cvut/kbss/termit/config/WebAppConfig.java
lukaskabc: Here, it was named `stringMessageConverter`, but it is `HttpMessageConverter`; since I was adding `MessageConverter`, I needed to make it different, and naming it `stringHttpMessageConverter` conflicts with beans from Spring autoconfiguration.

src/main/java/cz/cvut/kbss/termit/websocket/ResultWithHeaders.java
ledsoft: This is fine by me, just explain in the Javadoc why it should not be combined with other method-return-value handlers:)

src/main/java/cz/cvut/kbss/termit/config/WebSocketConfig.java
ledsoft: Yes, that is probably better. Or indicate more clearly that this comment concerns the following group of methods, it is quite unclear right now

src/main/java/cz/cvut/kbss/termit/config/WebAppConfig.java
ledsoft: Ok, leave it like this then.

src/main/java/cz/cvut/kbss/termit/websocket/handler/WebSocketExceptionHandler.java
lukaskabc: I agree, but method signatures for `@ExceptionHandler` from REST and `@MessageExceptionHandler` from WS cannot be merged.
So, do you want to have all those methods in a single class?

```Java
    @ExceptionHandler(PersistenceException.class)
    public ResponseEntity<ErrorInfo> persistenceException(HttpServletRequest request, PersistenceException e) {
        logException(e, request);
        return new ResponseEntity<>(errorInfo(request, e.getCause()), HttpStatus.INTERNAL_SERVER_ERROR);
    }

    @MessageExceptionHandler(PersistenceException.class)
    public ErrorInfo persistenceException(Message<?> message, PersistenceException e) {
        logException(e, message);
        return errorInfo(message, e.getCause());
    }

```

ledsoft: Ok, in that case leave it as is.

ledsoft (approved):
LGTM

PR #295, opened 2024-09-11
Method Throttling

### LINKED ISSUE #285: Performance when multiple users edit the same vocabulary
When multiple people edit the same vocabulary and the same time, TermIt's performance becomes quite bad (up to the point when proxy starts timing out on requests).

It seems the problem is that each edit of a term's label/definition leads to text analysis of all terms in the vocabulary, which is happening repeatedly and at the same time as terms are created and edited by the users. We should consider batching the analyses and possibly pushing notifications to client when the analysis is actually performed, and its results processed.

### LINKED ISSUE #287: Performance of vocabulary validation
When validating a moderately-sized vocabulary (ca 400 terms) with flat structure validation often times out on the proxy after 60 seconds. This is unacceptable.

lukaskabc:
This PR addresses performance issues from #287 and #285

# Identified problems
- **vocabulary validation** - was being executed on a request from the frontend. Results were cached, but the whole cache was evicted on any modification of any vocabulary. Vocabulary validation also requires synchronization as the validator is not thread-safe, blocking each thread assigned to handle the validation request until the previous validation is performed (This was probably also preventing the app from running out of memory when multiple validations would be processed concurrently). Validation is a time and resource-consuming task.
- **asynchronous term occurrence saving** - Many individual tasks might get scheduled during occurrence saving, but it's not even guaranteed when they will be saved; it could get to the point where the app was doing nothing but catching up on occurrence saving.
- **text analysis** - a similar problem as with validation, it's executed on term modification (or by a user on a file, for example); it is time and resource-consuming, and it results in async term occurrence saving. Text analysis of a term was executed before text analysis of all terms from the vocabulary, resulting in duplicated processing.
- **Single code deployment** - TermIt is deployed in a single core environment, eliminating the asynchronous processing benefits.

# New features
## Throttle & Debounce
This PR introduces an option to throttle and debounce method calls.

method throttling & debouncing

> Do not review linked test cases before reading about return type support and throttled futures

The goal of method throttling & debouncing is to execute tasks **asynchronously** on a **fixed** thread pool merging often method calls into a **single task** execution with the **newest data** from the most recent method call [[test case](https://github.com/lukaskabc/termit/blob/85d2e4b6b736ce91c7d4cc94230cc4c53a1aed7e/src/test/java/cz/cvut/kbss/termit/util/throttle/ThrottleAspectTest.java#L280)]. Throttling ensures that if the method task were not executed in the last [X seconds](https://github.com/lukaskabc/termit/blob/85d2e4b6b736ce91c7d4cc94230cc4c53a1aed7e/src/main/java/cz/cvut/kbss/termit/util/Constants.java#L59), it would be scheduled for immediate execution [[test case](https://github.com/lukaskabc/termit/blob/85d2e4b6b736ce91c7d4cc94230cc4c53a1aed7e/src/test/java/cz/cvut/kbss/termit/util/throttle/ThrottleAspectTest.java#L256)]. Otherwise, its execution will be delayed (debounced) so that it can be merged with potential future calls (it guarantees that when no future call comes, the task will be executed with the data from the last call). Task execution also ensures that when a task is time-consuming and its execution is taking longer than the actual throttle interval, a new call to the throttled method won't result in the concurrent execution of the same task [[test case](https://github.com/lukaskabc/termit/blob/85d2e4b6b736ce91c7d4cc94230cc4c53a1aed7e/src/test/java/cz/cvut/kbss/termit/util/throttle/ThrottleAspectTest.java#L876)].
When a thread is already executing a throttled task, it will ignore any further throttling and will execute all methods synchronously [[test case](https://github.com/lukaskabc/termit/blob/85d2e4b6b736ce91c7d4cc94230cc4c53a1aed7e/src/test/java/cz/cvut/kbss/termit/util/throttle/ThrottleAspectTest.java#L533)].

Throttling & Debouncing is realized with the [Throttle](https://github.com/lukaskabc/termit/blob/85d2e4b6b736ce91c7d4cc94230cc4c53a1aed7e/src/main/java/cz/cvut/kbss/termit/util/throttle/Throttle.java) annotation which is handled by the [Throttle aspect](https://github.com/lukaskabc/termit/blob/85d2e4b6b736ce91c7d4cc94230cc4c53a1aed7e/src/main/java/cz/cvut/kbss/termit/util/throttle/ThrottleAspect.java). 
> Aspect is configured using Spring AOC XML syntax to not utilize AspectJ. Once AspectJ is removed from dependencies, it should be possible to replace XML configuration with annotations. Aspect is disabled for the test profile.

Throttle annotation supports methods with **void** return type out of the box.
The whole method is, in that case, considered a task that should be throttled, and the method itself will be executed asynchronously.

There is also support for methods returning a [Future](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/Future.html). However, the concrete returned object **MUST** be [ThrottledFuture](https://github.com/lukaskabc/termit/blob/85d2e4b6b736ce91c7d4cc94230cc4c53a1aed7e/src/main/java/cz/cvut/kbss/termit/util/throttle/ThrottledFuture.java). Otherwise the aspect will throw appropriate exception on the method call (there is no way to safely check that on application start).
When a method returns future (ThrottledFuture), the method itself will be executed synchronously allowing to prepare the task that should be throttled and also to provide a cached result which may be acquired by a caller method from [CacheableFuture](https://github.com/lukaskabc/termit/blob/85d2e4b6b736ce91c7d4cc94230cc4c53a1aed7e/src/main/java/cz/cvut/kbss/termit/util/throttle/CacheableFuture.java) interface before the actual future resolution. The actual task and cached result is then provided through the returned ThrottledFuture object.
An example can be seen in updated [result caching validator](https://github.com/lukaskabc/termit/blob/85d2e4b6b736ce91c7d4cc94230cc4c53a1aed7e/src/main/java/cz/cvut/kbss/termit/persistence/validation/ResultCachingValidator.java#L94) where the method `validate` will be executed synchronously by the caller thread, checking the cache state and returning already resolved future when the cache is not dirty, or returning the future with the time-consuming task `runValidation` method and providing the cached result. Method `runValidation` will be executed asynchronously.

Throttled future also implements a [chainable future](https://github.com/lukaskabc/termit/blob/85d2e4b6b736ce91c7d4cc94230cc4c53a1aed7e/src/main/java/cz/cvut/kbss/termit/util/throttle/ChainableFuture.java) interface, which allows to chain a task that will be executed once the future is resolved.
This, for example, allows the [WebSocket controller](https://github.com/lukaskabc/termit/blob/85d2e4b6b736ce91c7d4cc94230cc4c53a1aed7e/src/main/java/cz/cvut/kbss/termit/websocket/VocabularySocketController.java#L58) to respond with the cached result and set a task that will send a new result to the client once new data are available. This prevents the thread from being blocked while awaiting a future resolution.

Scheduling throttled futures also support their cancellation based on their **group**.  
This, for example, allows to cancel scheduling a task to analyze a definition of a single term while an analysis of all terms from the vocabulary is scheduled.

## Disadvantages
- a throttled method can't use the previous transaction - when the throttled method is called during a transaction, the throttled task is executed asynchronously without access to the original transaction. However, when the [Transactional](https://docs.spring.io/spring-framework/docs/6.1.12/javadoc-api/org/springframework/transaction/annotation/Transactional.html) annotation is present on the same method as the Throttled annotation, the task will be executed in transactional context.
> Unfortunately, I was not able to make a detection of active transaction context work. It might be a feature missing in Jopa (and TransactionSynchronizationManager), or I might just miss something; anyway, the explicit transactional annotation is required for the transaction to work.

## Long-running tasks
As the application will now run some time-consuming tasks in the background, it will push the status of such tasks to the clients via WebSocket, allowing to display information about the activity to the user.

Currently, it's only possible to name the throttled method by a constant. So, the user will know that there is a validation in progress but won't know which vocabulary is being validated. This might be changed by adding a new parameter for additional information (in addition to the name parameter).

# Changes
- Removed periodic task for clearing context scheduled for removal (in case of term occurrence in definitions, removal is fast and will be made synchronously, for files see next point)
- Concurrent saving and resolving of term occurrences for file analysis. A second thread will be started, one thread will start resolving occurrences, while the second thread will execute the removal of all current ones; after successful removal, it will start to save resolved occurrences from the first thread.
- Vocabulary validation will be performed asynchronously using throttling.
- New vocabulary validation results will be pushed to clients via WebSocket.
- Text analysis will be performed asynchronously using throttling.
- Clients will be notified about the end of text analysis
- The frontend is no longer in control when validation and text analysis are performed (unless triggered by an explicit button by a user). The backend will automatically execute these tasks when appropriate (when vocabulary is modified, etc.).
- The result caching validator was rewritten to provide dirty cached results through throttling and only mark the cache as dirty instead of deleting it. It will also mark only related cache as dirty and won't touch other non-related entries.
- Removed the option to disable vocabulary analysis as discussed in a meeting

# Requirements and notes
- It is expected that TermIt will be deployed in an environment with at least two cores available to benefit from asynchronous processing (more cores would be, of course, beneficial as we need to handle http, websocket, database and background tasks).

- The annotation is not prepared for [AOT](https://docs.spring.io/spring-framework/reference/core/aot.html), and a reflection processor registering runtime hints will need to be probably created if the support for AOT is added.


src/main/java/cz/cvut/kbss/termit/config/WebAppConfig.java
ledsoft, line 1: These changes seem completely superfluous and don't seem to have any real effect.

src/main/java/cz/cvut/kbss/termit/event/VocabularyContentModifiedEvent.java
```diff
@@ -17,7 +17,7 @@
  */
 package cz.cvut.kbss.termit.event;
 
-import org.springframework.context.ApplicationEvent;
+import org.jetbrains.annotations.NotNull;
```
ledsoft: Use Spring's `NonNull` instead of the JetBrains `NotNull`.

src/main/java/cz/cvut/kbss/termit/event/VocabularyCreatedEvent.java
```diff
@@ -17,14 +17,16 @@
  */
 package cz.cvut.kbss.termit.event;
 
-import org.springframework.context.ApplicationEvent;
+import org.jetbrains.annotations.NotNull;
```
ledsoft: DTTO, use Spring `NonNull`.

src/main/java/cz/cvut/kbss/termit/event/VocabularyEvent.java
```diff
@@ -0,0 +1,28 @@
+package cz.cvut.kbss.termit.event;
+
+import org.jetbrains.annotations.NotNull;
```
ledsoft: DTTO, use Spring `NonNull`

src/main/java/cz/cvut/kbss/termit/event/VocabularyFileTextAnalysisFinishedEvent.java
```diff
@@ -0,0 +1,23 @@
...
+
+import cz.cvut.kbss.termit.model.resource.File;
+import org.jetbrains.annotations.NotNull;
+
+import java.net.URI;
+
+/**
+ * Indicates that text analysis of a file was finished
+ */
+public class VocabularyFileTextAnalysisFinishedEvent extends VocabularyEvent {
```
ledsoft: I would prefer to call it `FileTextAnalysisFinishedEvent`, without the `Vocabulary` part, as it is not that important.

src/main/java/cz/cvut/kbss/termit/event/VocabularyTermDefinitionTextAnalysisFinishedEvent.java
```diff
@@ -0,0 +1,22 @@
...
+
+import cz.cvut.kbss.termit.model.AbstractTerm;
+import org.jetbrains.annotations.NotNull;
+
+import java.net.URI;
+
+/**
+ * Indicates that a text analysis of a term definition was finished
+ */
+public class VocabularyTermDefinitionTextAnalysisFinishedEvent extends VocabularyEvent {
```
ledsoft: Again, remove the `Vocabulary` part of the class name, `TermDefinition` is enough to communicate the meaning of the class.

src/main/java/cz/cvut/kbss/termit/event/VocabularyValidationFinishedEvent.java
```diff
@@ -0,0 +1,50 @@
...
+ * Indicates that validation for a set of vocabularies was finished.
+ */
+public class VocabularyValidationFinishedEvent extends VocabularyEvent {
+
+    /**
+     * Vocabulary closure of {@link #vocabularyIri}.
+     * IRIs of vocabularies that are imported by {@link #vocabularyIri} and were part of the validation.
+     */
+    @NotNull
+    @Unmodifiable
```
ledsoft: We don't use this annotation in TermIt, remove it. It is superfluous since you create an unmodifiable collection in the constructor anyway (which is by itself uncommon I would say, as typically you would prevent modification on component interfaces, not within the component itself).

src/main/java/cz/cvut/kbss/termit/event/VocabularyValidationFinishedEvent.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Indicates that validation for a set of vocabularies was finished.
+ */
+public class VocabularyValidationFinishedEvent extends VocabularyEvent {
+
+    /**
+     * Vocabulary closure of {@link #vocabularyIri}.
+     * IRIs of vocabularies that are imported by {@link #vocabularyIri} and were part of the validation.
+     */
+    @NotNull
```
ledsoft: Remove the annotation from here, I think it is useless on a field, null check (or annotation) should be on component interface - constructor in this case.

src/main/java/cz/cvut/kbss/termit/event/VocabularyValidationFinishedEvent.java
```diff
@@ -0,0 +1,50 @@
+package cz.cvut.kbss.termit.event;
+
+import cz.cvut.kbss.termit.model.validation.ValidationResult;
+import org.jetbrains.annotations.NotNull;
```
ledsoft: Use Spring's `NonNull`.

src/main/java/cz/cvut/kbss/termit/event/VocabularyWillBeRemovedEvent.java
```diff
@@ -1,21 +1,15 @@
 package cz.cvut.kbss.termit.event;
 
-import org.springframework.context.ApplicationEvent;
+import org.jetbrains.annotations.NotNull;
```
ledsoft: Use Spring's `NonNull`

src/main/java/cz/cvut/kbss/termit/persistence/dao/TermOccurrenceDao.java
```diff
@@ -258,12 +254,11 @@ public void removeAll(Asset<?> target) {
...
-        em.createNativeQuery("MOVE GRAPH ?g TO ?targetContext")
-          .setParameter("g", sourceContext)
-          .setParameter("targetContext", targetContext)
-          .executeUpdate();
-        contextRemover.scheduleForRemoval(targetContext);
+        LOG.debug("Removing all occurrences from {}", sourceContext);
+        em.createNativeQuery("DROP GRAPH ?context")
+                .setParameter("context", sourceContext)
+                .executeUpdate();
+        LOG.debug("Removed all occurrences from {}", sourceContext);
```
ledsoft: Use `Utils.uriToString` to stringify sourceContext.

src/main/java/cz/cvut/kbss/termit/persistence/dao/VocabularyDao.java
```diff
@@ -144,17 +145,17 @@ public Vocabulary getReference(URI id) {
     /**
      * Gets identifiers of all vocabularies imported by the specified vocabulary, including transitively imported ones.
      *
-     * @param entity Base vocabulary, whose imports should be retrieved
+     * @param vocabulary Base vocabulary, whose imports should be retrieved
```
ledsoft: Identifier of base vocabulary... Also the parameter should be called `vocabularyIri`.

src/main/java/cz/cvut/kbss/termit/persistence/validation/ResultCachingValidator.java
```diff
@@ -17,24 +17,35 @@
...
+import cz.cvut.kbss.termit.event.EvictCacheEvent;
+import cz.cvut.kbss.termit.event.VocabularyContentModifiedEvent;
+import cz.cvut.kbss.termit.event.VocabularyCreatedEvent;
+import cz.cvut.kbss.termit.event.VocabularyEvent;
+import cz.cvut.kbss.termit.exception.TermItException;
 import cz.cvut.kbss.termit.model.validation.ValidationResult;
+import cz.cvut.kbss.termit.util.throttle.Throttle;
+import cz.cvut.kbss.termit.util.throttle.ThrottledFuture;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
```
ledsoft: Again, don't use JetBrains annotations, use Spring `NonNull`. We don't use `Nullable`, it is assumed by default.

src/main/java/cz/cvut/kbss/termit/persistence/validation/ResultCachingValidator.java
```diff
@@ -43,22 +54,103 @@ public class ResultCachingValidator implements VocabularyContentValidator {
 
     private static final Logger LOG = LoggerFactory.getLogger(ResultCachingValidator.class);
 
-    private final Map<Collection<URI>, List<ValidationResult>> validationCache = new ConcurrentHashMap<>();
+    /**
+     * Map of origin vocabulary IRI to vocabulary iri closure of imported vocabularies.
+     * When the record is missing, the cache is considered as dirty.
+     */
+    private final Map<URI, @NotNull Collection<URI>> vocabularyClosure = new ConcurrentHashMap<>();
```
ledsoft: Again, we don't use `NotNull` (or rather `NonNull`) annotations on fields, just on component interface (setters, constructors).

src/main/java/cz/cvut/kbss/termit/persistence/validation/ResultCachingValidator.java
```diff
@@ -43,22 +54,103 @@ public class ResultCachingValidator implements VocabularyContentValidator {
...
     Validator getValidator() {
         return null;    // Will be replaced by Spring
     }
 
-    @EventListener
-    public void evictCache(VocabularyContentModified event) {
-        LOG.debug("Vocabulary content modified, evicting validation result cache.");
-        validationCache.clear();
+    @EventListener({VocabularyContentModifiedEvent.class, VocabularyCreatedEvent.class})
+    public void evictVocabularyCache(VocabularyEvent event) {
```
ledsoft: Name is misleading, the method does not evict anything.

src/main/java/cz/cvut/kbss/termit/rest/ResourceController.java
ledsoft, line 1: All changes except for two lines (one OpenAPI doc and the other log message removal) are just formatting changes

src/main/java/cz/cvut/kbss/termit/rest/TermController.java
ledsoft, line 1: Again, only useless formatting changes that make the PR bigger for no reason.

src/main/java/cz/cvut/kbss/termit/rest/VocabularyController.java
ledsoft, line 1: More useless formatting changes. I propose to remove the warning logs, leave these controller as they are and later discuss what to do with the explicit analysis endpoints. We can either remove them altogether, or move the to administration controller.

src/main/java/cz/cvut/kbss/termit/rest/handler/RestExceptionHandler.java
```diff
@@ -80,7 +82,10 @@ private static void logException(Throwable ex, HttpServletRequest request) {
     }
 
     private static void logException(String message, Throwable ex) {
-        LOG.error(message, ex);
+        // prevents from logging exceptions caused be broken connection with a client
```
ledsoft: "Prevents exceptions caused by broken connection with client from logging"

src/main/java/cz/cvut/kbss/termit/security/JwtUtils.java
ledsoft, line 1: Useless formatting changes.

src/main/java/cz/cvut/kbss/termit/service/document/SynchronousTermOccurrenceSaver.java
ledsoft, line 1: Since there is no longer an asynchronous term occurrence saver, we can probably rename this one to just `TermOccurrenceSaver`.

ledsoft: Yeah, and merge it with the `TermOccurrenceSaver` inteface.

src/main/java/cz/cvut/kbss/termit/service/document/html/HtmlSelectorGenerators.java
```diff
@@ -52,6 +52,6 @@ public HtmlSelectorGenerators(Configuration config) {
      * @return Set of generated selectors
      */
     public Set<Selector> generateSelectors(Element... elements) {
-        return generators.stream().map(g -> g.generateSelector(elements)).collect(Collectors.toSet());
+        return generators.parallelStream().map(g -> g.generateSelector(elements)).collect(Collectors.toSet());
```
ledsoft: It is generally discouraged to use parallelStream in web application environments because they by default use the same threadpool. Also, do you have evidence that using a parallelStream in this particular case leads to a noticeable speedup? If not, please revert the change.

src/main/java/cz/cvut/kbss/termit/service/document/html/TextPositionSelectorGenerator.java
```diff
@@ -47,15 +51,31 @@ public TextPositionSelector generateSelector(Element... elements) {
...
+    /**
+     * @see SelectorGenerator#extractNodeText(Iterable)
+     * @see Element#wholeText()
+     * @see TextNode#getWholeText()
+     */
     private int resolveStartPosition(Element element) {
         final Elements ancestors = element.parents();
         Element previous = element;
-        int counter = 0;
+        AtomicInteger counter = new AtomicInteger();
```
ledsoft: Why are you using an `AtomicInteger`? I don't see any way multiple threads could access the counter, so I feel the change brings nothing.

ledsoft: Also, if this (and the rest of the changes in this file) is related to the parallelStream usage in HtmlSelectorGenerators, see my comment there.

src/main/java/cz/cvut/kbss/termit/util/Constants.java
```diff
@@ -46,6 +48,22 @@ public class Constants {
...
+     * @see cz.cvut.kbss.termit.util.throttle.Throttle
+     * @see cz.cvut.kbss.termit.util.throttle.ThrottleAspect
+     */
+    public static final Duration THROTTLE_THRESHOLD = Duration.ofSeconds(10);
+
+    /**
+     * After how much time, should complete futures be discarded.
+     * @see ThrottleAspect#clearOldFutures()
+     */
+    public static final Duration THROTTLE_DISCARD_THRESHOLD = Duration.ofMinutes(1);
```
ledsoft: These should be configurable, too.

src/main/java/cz/cvut/kbss/termit/util/Pair.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public V getSecond() {
+        return second;
+    }
+
+
+    /**
+     * First compares the first value, if they are equal, compares the second value.
+     */
+    public static class Comparable<T extends java.lang.Comparable<T>, V extends java.lang.Comparable<V>>
```
ledsoft: `ComparablePair` perhaps?

src/test/java/cz/cvut/kbss/termit/service/document/AnnotationGeneratorTest.java
ledsoft, line 1: Useless formatting changes.

src/test/java/cz/cvut/kbss/termit/rest/ResourceControllerTest.java
ledsoft, line 1: Again, I see only formatting changes.

src/main/java/cz/cvut/kbss/termit/model/Vocabulary.java
```diff
@@ -236,13 +237,23 @@ public int hashCode() {
...
     public String toString() {
-        return "Vocabulary{" +
-                getLabel() +
-                " " + Utils.uriToString(getUri()) +
-                ", glossary=" + glossary +
-                (importedVocabularies != null ?
-                 ", importedVocabularies = [" + importedVocabularies.stream().map(Utils::uriToString).collect(
-                         Collectors.joining(", ")) + "]" : "") +
-                '}';
+        final StringBuilder builder = new StringBuilder();
```
ledsoft: Don't use `StringBuilder` unless constructing some complex string (or in a loop), the JVM is able to optimize string concatenation as necessary.

ledsoft (changes requested):
Haven't tried, yet so these are purely based on reviewing the code here.

A bunch of inline comments, plus these general remarks:
- Replace JetBrains annotations (particularly `NotNull`) with their Spring equivalents (`NonNull`). I stopped pointing that out after a couple of files. We don't use `Nullable`, it is assumed by default. I also don't see any reason for using these annotations on fields, use them only on public methods
- There are several files containing formatting changes only. This makes the already large PR even larger
- I also have a question regarding the necessity of using XML-configured aspect, but let's discuss that on our meeting

src/main/java/cz/cvut/kbss/termit/rest/ResourceController.java
lukaskabc: keeping without change as discussed on our meeting

src/main/java/cz/cvut/kbss/termit/config/WebAppConfig.java
lukaskabc: keeping without change as discussed on our meeting

src/main/java/cz/cvut/kbss/termit/rest/TermController.java
lukaskabc: keeping without change as discussed on our meeting

src/main/java/cz/cvut/kbss/termit/security/JwtUtils.java
lukaskabc: keeping without change as discussed on our meeting

src/test/java/cz/cvut/kbss/termit/service/document/AnnotationGeneratorTest.java
lukaskabc: keeping without change as discussed on our meeting

src/test/java/cz/cvut/kbss/termit/rest/ResourceControllerTest.java
lukaskabc: keeping without change as discussed on our meeting

src/main/java/cz/cvut/kbss/termit/service/document/html/TextPositionSelectorGenerator.java
lukaskabc: The atomic reference was required for access from the lambda function. However, I reviewed the changes and reverted them as they didn't have the desired effect. It was an attempt to eliminate the construction of a whole string to retrieve its length.

ledsoft (approved):
LGTM, works nicely. I added a small fix for situations where terms are related to each other, as it was throwing exception due to the analysis of the whole vocabulary running in the same transaction.

PR #305, opened 2024-10-13
[Enhancement kbss-cvut/termit-ui#519] Term history in vocabulary

lukaskabc:
resolves kbss-cvut/termit-ui#519

Adds `/vocabularies/{localName}/history-of-content/detail` endpoint, which returns change records of vocabulary terms.


ledsoft (approved):
LGTM

PR #322, opened 2024-12-10
[Enhancement kbss-cvut/termit-ui#535] Ability to clear validation/text analysis queue

lukaskabc:
Add maintenance endpoint that allows an administrator to clear the long-running tasks queue.  
All non-running throttled futures are cancelled and non-running tasks are removed from the long-running task registry.


ledsoft (approved):
LGTM

PR #325, opened 2025-01-29
[Enhancement kbss-cvut/termit-ui#481] Allow disabling public view of all vocabularies/ vocabulary

lukaskabc:
Introduces a new user role - anonymous user, representing a not-logged-in user.

Access Control Lists / Records are now being validated, enforcing the presence of records for anonymous, restricted, and full user roles.
Validation also ensures that an access control record:
- Does not grant greater access level than read to the anonymous user role
- Does not grant security access level to the restricted user role
- Does not grant security access level to a user group

Until now, these rules were only controlled by FE.

Init service `VocabularyAnonymousAccessControlListGenerator` was introduced for *data migration* to generate missing access control records for anonymous user role.

Resolves kbss-cvut/termit-ui#481

### DIFF
--- a/ontology/termit-glosář.ttl
+++ b/ontology/termit-glosář.ttl
@@ -384,6 +384,17 @@ termit-pojem:omezený-uživatel-termitu
         <http://www.w3.org/2004/02/skos/core#scopeNote>
                 "He or she can list and view vocabularies and terms. He or she can also comment on terms."@en , "Může prohlížet slovníky a pojmy. Pojmy může rovněž komentovat."@cs .
 
+termit-pojem:anonymní-uživatel-termitu
+        a       <http://www.w3.org/2004/02/skos/core#Concept> ;
+        <http://www.w3.org/2004/02/skos/core#broader>
+                termit-pojem:uživatel-termitu , <https://slovník.gov.cz/základní/pojem/typ-objektu> ;
+        <http://www.w3.org/2004/02/skos/core#inScheme>
+                termit:glosář ;
+        <http://www.w3.org/2004/02/skos/core#prefLabel>
+                "Anonymous"@en , "Anonymní"@cs ;
+        <http://www.w3.org/2004/02/skos/core#scopeNote>
+        "A not logged in user"@en , "Nepřihlášený uživatel"@cs .
+
 termit-pojem:má-suffix-text-quote
         a       <http://www.w3.org/2004/02/skos/core#Concept> ;
         <http://www.w3.org/2004/02/skos/core#broader>
--- a/ontology/termit-glosář.ttl
+++ b/ontology/termit-glosář.ttl
@@ -390,8 +390,8 @@ termit-pojem:anonymní-uživatel-termitu
                 termit-pojem:uživatel-termitu , <https://slovník.gov.cz/základní/pojem/typ-objektu> ;
         <http://www.w3.org/2004/02/skos/core#inScheme>
                 termit:glosář ;
-        <http://www.w3.org/2004/02/skos/core#prefLabel>
-                "Anonymous"@en , "Anonymní"@cs ;
+        <http://www.w3.org/2004/02/skos/core#prefLabel> 
+                "Anonymous"@en , "Anonym"@cs ;
         <http://www.w3.org/2004/02/skos/core#scopeNote>
         "A not logged in user"@en , "Nepřihlášený uživatel"@cs .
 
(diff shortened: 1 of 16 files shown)


src/main/java/cz/cvut/kbss/termit/service/init/VocabularyAnonymousAccessControlListGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+     * {@link AccessLevel#NONE AccessLevel#NONE} is assigned when the restricted user has no access,
+     * otherwise {@link AccessLevel#READ AccessLevel#READ} is assigned for the anonymous user.
+     * <p>
+     * This method is asynchronous to prevent slowing down the system startup.
+     */
+    @Async
+    @Transactional
+    public void generateMissingAccessControlLists() {
+        LOG.debug("Generating missing vocabulary access control records for anonymous users.");
+        final UserRole anonymousRole = userRoleRepositoryService.findAll().stream()
```
ledsoft: I think you can just use the anonymous role's IRI constant instead of finding all and then iterating through them.

src/main/java/cz/cvut/kbss/termit/service/repository/RepositoryAccessControlListService.java
```diff
@@ -179,25 +182,120 @@ public void removeRecord(AccessControlList acl, AccessControlRecord<?> record) {
...
+     * </ul>
+     *
+     * @param controlRecord Access control record to validate
+     * @throws UnsupportedOperationException if the record is not valid
+     */
+    public void validate(AccessControlRecord<?> controlRecord) {
+        if (controlRecord.getHolder() instanceof UserRole role) {
+            // check that the anonymous user does not have greater access level than READ
+            if (isAnonymous(role) && controlRecord.getAccessLevel().compareTo(AccessLevel.READ) > 0) {
+                throw new UnsupportedOperationException("Access control record for anonymous user cannot grant greater access level then READ.");
```
ledsoft: "then" -> "than"

ontology/termit-glosář.ttl
```diff
@@ -384,6 +384,17 @@ termit-pojem:omezený-uživatel-termitu
...
                 "He or she can list and view vocabularies and terms. He or she can also comment on terms."@en , "Může prohlížet slovníky a pojmy. Pojmy může rovněž komentovat."@cs .
 
+termit-pojem:anonymní-uživatel-termitu
+        a       <http://www.w3.org/2004/02/skos/core#Concept> ;
+        <http://www.w3.org/2004/02/skos/core#broader>
+                termit-pojem:uživatel-termitu , <https://slovník.gov.cz/základní/pojem/typ-objektu> ;
+        <http://www.w3.org/2004/02/skos/core#inScheme>
+                termit:glosář ;
+        <http://www.w3.org/2004/02/skos/core#prefLabel>
+                "Anonymous"@en , "Anonymní"@cs ;
```
ledsoft: "Anonymní" -> "Anonym"

src/main/java/cz/cvut/kbss/termit/service/init/VocabularyAnonymousAccessControlListGenerator.java
ledsoft: I realize this will require adding a find method to UserRoleService and DAO, but I think it makes sense in this case.

ledsoft (commented):
Just a couple of nitpicky comments. Otherwise looks good and works nicely.

src/main/java/cz/cvut/kbss/termit/persistence/dao/UserRoleDao.java
```diff
@@ -35,4 +36,9 @@ public UserRoleDao(EntityManager em) {
     public List<UserRole> findAll() {
         return em.createQuery("SELECT r FROM " + UserRole.class.getSimpleName() + " r", UserRole.class).getResultList();
     }
+
+    public UserRole find(cz.cvut.kbss.termit.security.model.UserRole userRole) {
```
ledsoft: Just em `em.find`, no query is needed.

ledsoft: In addition, to be consistent with `find` in other DAOs, I suggest returning an `Optional` and handling missing value in the service.

src/main/java/cz/cvut/kbss/termit/service/repository/UserRoleRepositoryService.java
```diff
@@ -37,4 +37,8 @@ public UserRoleRepositoryService(UserRoleDao dao) {
     public List<UserRole> findAll() {
         return dao.findAll();
     }
+
+    public UserRole find(cz.cvut.kbss.termit.security.model.UserRole userRole) {
```
ledsoft: Check that the result returned by DAO is not empty (see my comment in DAO). See other services and their `findRequired` methods.

ledsoft (changes requested):

ledsoft (approved):
LGTM

PR #328, opened 2025-02-05
[Ref #318] Remove dependency on AspectJ

### LINKED ISSUE #318: Remove dependency on AspectJ
Now that JOPA does not require AspectJ, TermIt depends on AspectJ only because some cross-cutting behavior is implemented by aspects. We should replace this by other suitable means (e.g., Spring application events) and remove the dependency on AspectJ.

lukaskabc:
resolves #318 

Removed:
- aspectjrt
- spring-aspects
- aspectj-maven-plugin

Added:
- spring-aop (for ThrottleAspect)

Maven placeholders in `application.properties` were added into quotes, which allows us to build and run the application in IDE without processing resources with Maven.

Removed `@Configured` from entities (Term, Vocabulary) for autowiring configuration.

Accessors for `PrimaryLabel` were refactored to  
`String getLabel(String language)` and `void setLabel(String language, String label)`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,7 +163,7 @@
         </dependency>
         <dependency>
             <groupId>org.springframework</groupId>
-            <artifactId>spring-aspects</artifactId>
+            <artifactId>spring-aop</artifactId>
         </dependency>
         <dependency>
             <groupId>org.springframework.retry</groupId>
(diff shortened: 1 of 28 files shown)


ledsoft (approved):
LGTM, nice work!

PR #330, opened 2025-02-21
[Perf #327] Improve performance of analysis of all vocabulary terms

### LINKED ISSUE #327: Improve performance of analysis of all vocabulary terms
When a term's label or definition changes, it triggers text analysis of all terms in the same vocabulary. This can be quite time consuming when a vocabulary is large. As each term is analyzed separately, we should consider trying to batch the analysis, for example by joining the definitions into a single string that could be analyzed in one go and then separate the parts again in post-processing.

lukaskabc:
Resolves #327 

Text analysis of all terms inside a vocabulary will now merge definitions of all terms and invoke text analysis only once (per context).

Each merged term definition is wrapped with `termdefinition` tag.
```html
<termdefinition id="http://resource.identifier/term">
This is term definition
</termdefinition>
```

The result of text analysis is then processed with regex, extracting individual definitions and term IDs.  
Each annotated definition is then passed along with the term to the annotation generator individually.

Loading all the terms from a vocabulary will still take a decent amount of time.

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
+++ b/src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
@@ -329,6 +329,8 @@ public class TermDao extends BaseAssetDao<Term> implements SnapshotProvider<Term
     /**
      * Finds all terms in the specified vocabulary, regardless of their position in the term hierarchy.
      * Filters terms that have label and definition in the instance language.
+     * <p>
+     * Terms are loaded <b>without</b> their subterms.
      *
      * @param vocabulary Vocabulary whose terms to retrieve. A reference is sufficient
      * @return List of vocabulary term DTOs ordered by label
--- a/src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
+++ b/src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
@@ -326,6 +326,40 @@ public class TermDao extends BaseAssetDao<Term> implements SnapshotProvider<Term
         }
     }
 
+    /**
+     * Finds all terms in the specified vocabulary, regardless of their position in the term hierarchy.
+     * Filters terms that have label and definition in the instance language.
+     *
+     * @param vocabulary Vocabulary whose terms to retrieve. A reference is sufficient
+     * @return List of vocabulary term DTOs ordered by label
+     */
+    public List<TermDto> findAllWithDefinition(Vocabulary vocabulary) {
+        Objects.requireNonNull(vocabulary);
+        try {
+            return em.createNativeQuery("SELECT DISTINCT ?term WHERE {" +
+                                     "GRAPH ?context { " +
+                                     "?term a ?type ;" +
+                                     "?hasLabel ?label ;" +
+                                     "?hasDefinition ?definition ;" +
+                                     "FILTER (lang(?label) = ?labelLang) ." +
+                                     "FILTER (lang(?definition) = ?labelLang) ." +
+                                     "}" +
+                                     "?term ?inVocabulary ?vocabulary ." +
+                                     " } ORDER BY " + orderSentence("?label"),
+                             TermDto.class)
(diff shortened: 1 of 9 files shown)


src/main/java/cz/cvut/kbss/termit/service/business/VocabularyService.java
```diff
@@ -361,16 +362,22 @@ public List<AbstractChangeRecord> getDetailedHistoryOfContent(Vocabulary vocabul
               group = "T(ThrottleGroupProvider).getTextAnalysisVocabularyAllTerms(#vocabulary.getUri())",
               name = "allTermsVocabularyAnalysis")
     @PreAuthorize("@vocabularyAuthorizationService.canModify(#vocabulary)")
-    public void runTextAnalysisOnAllTerms(Vocabulary vocabulary) {
-        vocabulary = findRequired(vocabulary.getUri()); // required when throttling for persistent context
+    public void runTextAnalysisOnAllTerms(URI vocabualryUri) {
```
ledsoft: Typo in the variable name.

src/main/java/cz/cvut/kbss/termit/service/business/VocabularyService.java
```diff
@@ -361,16 +362,22 @@ public List<AbstractChangeRecord> getDetailedHistoryOfContent(Vocabulary vocabul
               group = "T(ThrottleGroupProvider).getTextAnalysisVocabularyAllTerms(#vocabulary.getUri())",
               name = "allTermsVocabularyAnalysis")
     @PreAuthorize("@vocabularyAuthorizationService.canModify(#vocabulary)")
```
ledsoft: Need to update the reference to the parameter here as well, otherwise it leads to NPX in the vocabulary authorization service.

src/main/java/cz/cvut/kbss/termit/service/business/VocabularyService.java
```diff
@@ -361,16 +362,22 @@ public List<AbstractChangeRecord> getDetailedHistoryOfContent(Vocabulary vocabul
               group = "T(ThrottleGroupProvider).getTextAnalysisVocabularyAllTerms(#vocabulary.getUri())",
               name = "allTermsVocabularyAnalysis")
```
ledsoft: Throttle needs to be also updated based on the change of the method signature.

ledsoft (changes requested):
There are some issues with the signature changes of `VocabularyService.runTextAnalysisOnAllTerms` (see the inline comments).

Otherwise great job, on a vocabulary with 798 terms the analysis time went down from 8min 43s to 30s (disabled cache)!

src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
```diff
@@ -326,6 +326,40 @@ public List<TermDto> findAll(Vocabulary vocabulary) {
         }
     }
 
+    /**
+     * Finds all terms in the specified vocabulary, regardless of their position in the term hierarchy.
+     * Filters terms that have label and definition in the instance language.
+     *
+     * @param vocabulary Vocabulary whose terms to retrieve. A reference is sufficient
+     * @return List of vocabulary term DTOs ordered by label
+     */
```
ledsoft: Could you also add information that the results do not contain sub terms?

src/main/java/cz/cvut/kbss/termit/service/business/TermService.java
```diff
@@ -138,6 +139,18 @@ public List<TermDto> findAll(Vocabulary vocabulary) {
         return repositoryService.findAll(vocabulary);
     }
 
+    /**
+     * Finds all terms in the specified vocabulary, regardless of their position in the term hierarchy.
```
ledsoft, line 143: Again, could you add information that the result terms do not have subterms?

ledsoft (commented):
Looks good and works great, just the two minor comments, then we can merge.

ledsoft (approved):
LGTM, nice work!

PR #350, opened 2025-08-08
[kbss-cvut/termit-ui#430] Podpora specifikace jazyka při FTS hledání

### LINKED ISSUE #430: Podpora specifikace jazyka při FTS hledání
# Popis
Aktuálně TermIt podporuje jenom jeden jazyk pro FTS vyhledavání ve slovnících. Bylo by fajn, kdyby si uživatel mohl volitelně vybrat i jiný jazyk než hlavní jazyk instance.

# Využití
V případě, že mám vícejazyčné slovníky, toto vyhledávání se hodí když chci vyhledávat v jiném než hlavním jazyku deploymentu.

# Užitečnost
3. bylo by hezké ji mít, ale bez problémů se bez ní obejdu
****

lukaskabc:
Resolves kbss-cvut/termit-ui#430

Adds Lucene connector management to Termit.
I investigated an option of removing the language filter from existing Lucene indexes, which results in merging all multilingual values of a field into one, which is then indexed. Unfortunately, in that case, it is not possible to search only in a specific language.

- Termit will, on each start (and every day at 03:00 AM), reinitialize missing Lucene indexes and drop unused ones.
- The process can also be triggered with the management endpoint
- For every language used on any indexed field, two Lucene indexes will be created - Label index and Definition Index
    - I think it is reasonable to assume that if a label exists in a language, a definition will be added in the language as well, and vice versa
- If some language is no longer used, existing indexes for it will be dropped from the database
- Termit will check Lucene connectors URI and will only manage those matching specified patterns
- Lucene connector options in the database are also verified for each existing instance, and the index is recreated with new options if a mismatch is found (they will be auto-updated if the definition in Termit changes)

- Full Text Search can now be performed in all languages or in a specific language only


src/main/java/cz/cvut/kbss/termit/rest/SearchController.java
```diff
@@ -68,8 +68,10 @@ public SearchController(IdentifierResolver idResolver, Configuration config, Sea
     @PreAuthorize("permitAll()")
     @GetMapping(value = "/fts", produces = {MediaType.APPLICATION_JSON_VALUE, JsonLd.MEDIA_TYPE})
     public List<FullTextSearchResult> fullTextSearch(@Parameter(description = "Search string.")
-                                                     @RequestParam(name = "searchString") String searchString) {
-        return searchService.fullTextSearch(searchString);
+                                                     @RequestParam(name = "searchString") String searchString,
+                                                     @Parameter(description = "Language of the search string")
```
ledsoft: This is a bit misleading. It is not only the language of the search string, but also the language of the search itself. I would probably rephrase it to something as simple as "Search language"

ledsoft: The same applies to the method below.

ledsoft: Also the description should mention that not specifying it will match all languages (as I found out in the `SearchService` Javadoc).

src/main/java/cz/cvut/kbss/termit/util/Configuration.java
```diff
@@ -361,6 +362,18 @@ public static class Persistence {
...
+         */
+        @NotEmpty
+        String luceneLabelIndexPrefix = "termit_labelIndex_";
+
+        /**
+         * Prefix used for naming lucene connectors indexing definitions.
+         */
+        @NotEmpty
+        String luceneDefcomIndexPrefix = "termit_defcomIndex_";
+
```
ledsoft: Let's remove this. I don't think it is necessary to make the connector prefixes configurable. It just increases the complexity of the already complex configuration of TermIt.

src/test/java/cz/cvut/kbss/termit/rest/SearchControllerTest.java
```diff
@@ -45,6 +44,7 @@
 import java.util.List;
 import java.util.Set;
 
+import static cz.cvut.kbss.termit.environment.Environment.LANGUAGE;
```
ledsoft: Please do not use static imports of constants, it is [advised against](https://docs.oracle.com/javase/1.5.0/docs/guide/language/static-import.html).

src/test/java/cz/cvut/kbss/termit/service/business/SearchServiceTest.java
```diff
@@ -43,6 +42,7 @@
 import java.util.List;
 import java.util.Set;
 
+import static cz.cvut.kbss.termit.environment.Environment.LANGUAGE;
```
ledsoft: Please do not use static imports of constants, it is [advised against](https://docs.oracle.com/javase/1.5.0/docs/guide/language/static-import.html).

src/test/java/cz/cvut/kbss/termit/rest/SearchControllerTest.java
ledsoft, line 1: Could you add a test for specifying the search language parameter in request?

src/test/java/cz/cvut/kbss/termit/persistence/dao/lucene/LuceneSearchDaoTest.java
```diff
@@ -60,21 +60,21 @@ class LuceneSearchDaoTest {
     @BeforeEach
     void setUp() {
         this.sut = new LuceneSearchDao(emMock, configMock);
+        sut.loadQueries();
```
ledsoft, line 57: Is this the only reason for making the `loadQueries` method public? In that case package-private should be sufficient, right?

src/main/java/cz/cvut/kbss/termit/service/init/lucene/LuceneConnectorInitializerImpl.java
```diff
@@ -0,0 +1,382 @@
...
+*/
+
+/**
+ * Finds all languages used by any object in database identified by an indexed predicate
+ * and ensures that required lucene connectors are created in the database with up-to-date options for each language.
+ * Remaining lucene connectors that are matching prefixes of {@link #requiredConnectors} are dropped.
+ */
+@Service
+@Profile("lucene")
+public class LuceneConnectorInitializerImpl implements LuceneConnectorInitializer {
```
ledsoft: I suggest renaming to `GraphDBLuceneConnectorInitializer`

src/main/java/cz/cvut/kbss/termit/service/init/lucene/LuceneConnectorInitializerImpl.java
```diff
@@ -0,0 +1,382 @@
...
+    public static final String LUCENE_INSTANCE_NS = "http://www.ontotext.com/connectors/lucene/instance#";
+    public static final URI LUCENE_LIST_CONNECTORS = URI.create("http://www.ontotext.com/connectors/lucene#listConnectors");
+    public static final URI LUCENE_LIST_OPTION_VALUES = URI.create("http://www.ontotext.com/connectors/lucene#listOptionValues");
+    public static final URI LUCENE_DROP_CONNECTOR = URI.create("http://www.ontotext.com/connectors/lucene#dropConnector");
+    public static final URI LUCENE_CREATE_CONNECTOR = URI.create("http://www.ontotext.com/connectors/lucene#createConnector");
```
ledsoft: Why are these public? I don't see them used anywhere in the project.

ledsoft (changes requested):
See the inline comments

src/main/java/cz/cvut/kbss/termit/service/init/lucene/LuceneConnectorInitializerImpl.java
lukaskabc: They are used in tests -> changed to package private

ledsoft (commented):
LGTM, one more thing: I don't think it makes sense to create connectors for languages for which there is no analyzer...

lukaskabc:
That would cause the search to not work at all for the filtered language.
Currently, when the analyzer does not exist, based on [GraphDB docs](https://graphdb.ontotext.com/documentation/10.0/lucene-graphdb-connector.html#overview-and-features), it will use the [StandardAnalyzer](https://github.com/apache/lucene/blob/41fcc22c76f85430bcbf10f3fdf78cbcc0e30ad5/lucene/core/src/java/org/apache/lucene/analysis/standard/StandardAnalyzer.java#L34). The same analyzer is used for indexing without a language filter.

ledsoft:
> That would cause the search to not work at all for the filtered language. Currently, when the analyzer does not exist, based on [GraphDB docs](https://graphdb.ontotext.com/documentation/10.0/lucene-graphdb-connector.html#overview-and-features), it will use the [StandardAnalyzer](https://github.com/apache/lucene/blob/41fcc22c76f85430bcbf10f3fdf78cbcc0e30ad5/lucene/core/src/java/org/apache/lucene/analysis/standard/StandardAnalyzer.java#L34). The same analyzer is used for indexing without a language filter.

Ok, I see. I think my comment was biased by using my own test instance where I was creating lots of translations with often non-existent languages. Never mind, merging the PR. Good work!

ledsoft (approved):
LGTM

PR #353, opened 2025-08-15
[Refactoring #351] Remove SearchDao

### LINKED ISSUE #351: Remove SearchDao
Remove SearchDao and replace it with LuceneSearchDao since Termit can only be used with GraphDB.

lukaskabc:
resolves #351


src/main/java/cz/cvut/kbss/termit/persistence/dao/FacetedSearchDao.java
ledsoft, line 1: Does a standalone `FacetedSearchDao` make sense? It seems to me the `LuceneSearchDao` subclass it used everywhere. What about merging both into a `SearchDao` and, if necessary, create interfaces that can be mocked for tests...

lukaskabc: I kept it separated, basically only for tests. I can merge it, but I would not merge the two classes with tests.

ledsoft: Separate test classes are certainly fine by me.

lukaskabc: ok, dao merged

src/main/java/cz/cvut/kbss/termit/persistence/dao/LuceneSearchDao.java
```diff
@@ -19,41 +19,67 @@
...
+ * Search data access object using Lucene-based repositories. These support rich search strings with wildcards and
  * operators.
  * <p>
  * This DAO automatically adds a wildcard to the last token in the search string, so that results for incomplete words
  * are returned as well.
  */
 @Repository
-@Profile("lucene")  // Corresponds to a profile set in pom.xml
-public class LuceneSearchDao extends SearchDao {
+public class LuceneSearchDao {
```
ledsoft: Can we call it `SearchDao` now? I don't think it is necessary to indicate it being based on Lucene indexes with class name (now that there is no alternative from which we would need to differentiate it).

lukaskabc: ok, dao renamed

ledsoft (approved):
LGTM

PR #354, opened 2025-08-18
[kbss-cvut/termit-ui#659] Restrict full-text search language selection to available languages

### LINKED ISSUE #659: Restrict full-text search language selection to available languages
The full-text search language selector (#430) currently allows selecting any of the languages supported by the ISO package we use. However, in reality only a small subset of languages is actually used in the data. It would make sense to restrict the selection only to languages actually present in the data.

lukaskabc:
resolves kbss-cvut/termit-ui#659
- Caches indexed languages in `GraphDBLuceneConnectorInitializer`
- Indexed languages are now included in the configuration endpoint available for the FE

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/service/config/ConfigurationProvider.java
+++ b/src/main/java/cz/cvut/kbss/termit/service/config/ConfigurationProvider.java
@@ -18,6 +18,7 @@
 package cz.cvut.kbss.termit.service.config;
 
 import cz.cvut.kbss.termit.dto.ConfigurationDto;
+import cz.cvut.kbss.termit.service.init.lucene.GraphDBLuceneConnectorInitializer;
 import cz.cvut.kbss.termit.service.repository.UserRoleRepositoryService;
 import cz.cvut.kbss.termit.util.Configuration;
 import cz.cvut.kbss.termit.util.Vocabulary;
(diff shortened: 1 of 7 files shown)


src/main/java/cz/cvut/kbss/termit/service/config/ConfigurationProvider.java
```diff
@@ -60,6 +65,9 @@ public ConfigurationDto getConfiguration() {
         result.setMaxFileUploadSize(maxFileUploadSize);
         result.setVersionSeparator(config.getNamespace().getSnapshot().getSeparator());
         result.setModelingToolUrl(config.getModelingToolUrl());
+        if (luceneConnectorInitializer != null) {
```
ledsoft: What if we removed the if, make the connector initializer required (the only circumstance there is no instance of the initializer is with the `test` profile) and provide a mock instance in tests? Maybe introduce an interface (`LuceneIndexInitializer`) for easier mocking...

ledsoft (commented):
LGTM, one proposal for a bit of refactoring.

src/main/java/cz/cvut/kbss/termit/service/config/ConfigurationProvider.java
lukaskabc: Introduced `IndexedLanguagesProvider` interface for getting the list of indexed languages. Mocked with an empty Set in the test environment.

ledsoft (approved):
LGTM, nice work!

PR #362, opened 2025-10-27
[Enhancement #338] Limit the frequency of document file backup creation

### LINKED ISSUE #338: Make file backups more space-efficient
Since we backup Files on every changes, active use of the annotator tends to massively increase the number of these files. Especially when approving suggested occurrences creates new backups as well. This may cause problems for server backups because the Docker volumes TermIt uses are full of File backups.

We should consider making the file handling more space-efficient, for example by:
- reducing the number of backups, perhaps by making simple changes (like approving an occurrence) directly in place instead of making a copy of the file
- Compressing the backup files

lukaskabc:
- Resource controller now accepts an optional reason for saving the content of a file (defaults to `unknown`)
- File backup is now only created for file `reupload`, `new occurrence` or `remove occurrence` reasons
- Added `modified` and `last backup` timestamps to the document file entity
- Added a scheduled task at 03:00, which creates a backup for all files that were modified since the last backup

### DIFF
--- a/ontology/termit-model.ttl
+++ b/ontology/termit-model.ttl
@@ -358,6 +358,7 @@ termit-pojem:vlastní-atribut
 
 termit-pojem:má-datum-a-čas-poslední-zálohy
         a                   owl:DatatypeProperty , <https://slovník.gov.cz/základní/pojem/typ-vztahu> ;
-        rdfs:subPropertyOf  <http://onto.fel.cvut.cz/ontologies/slovník/agendový/popis-dat/pojem/má-datum-a-čas-modifikace> .
+        rdfs:subPropertyOf  <http://onto.fel.cvut.cz/ontologies/slovník/agendový/popis-dat/pojem/má-datum-a-čas-modifikace> ;
+        rdfs:range          xsd:dateTime .
 
 
--- a/ontology/termit-model.ttl
+++ b/ontology/termit-model.ttl
@@ -356,4 +356,8 @@ termit-pojem:vlastní-atribut
         a                owl:Class ;
         rdfs:subClassOf rdf:Property .
 
+termit-pojem:má-datum-a-čas-poslední-zálohy
+        a                   owl:DatatypeProperty , <https://slovník.gov.cz/základní/pojem/typ-vztahu> ;
+        rdfs:subPropertyOf  <http://onto.fel.cvut.cz/ontologies/slovník/agendový/popis-dat/pojem/má-datum-a-čas-modifikace> .
+
 
--- a/src/main/java/cz/cvut/kbss/termit/service/business/ResourceService.java
+++ b/src/main/java/cz/cvut/kbss/termit/service/business/ResourceService.java
@@ -37,7 +37,6 @@ import cz.cvut.kbss.termit.service.changetracking.ChangeRecordProvider;
 import cz.cvut.kbss.termit.service.document.DocumentManager;
 import cz.cvut.kbss.termit.service.document.ResourceRetrievalSpecification;
 import cz.cvut.kbss.termit.service.document.TextAnalysisService;
-import cz.cvut.kbss.termit.service.document.backup.DocumentBackupManager;
 import cz.cvut.kbss.termit.service.document.html.UnconfirmedTermOccurrenceRemover;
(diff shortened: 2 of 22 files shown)


ontology/termit-model.ttl
```diff
@@ -356,4 +356,8 @@ termit-pojem:vlastní-atribut
         a                owl:Class ;
         rdfs:subClassOf rdf:Property .
 
+termit-pojem:má-datum-a-čas-poslední-zálohy
+        a                   owl:DatatypeProperty , <https://slovník.gov.cz/základní/pojem/typ-vztahu> ;
+        rdfs:subPropertyOf  <http://onto.fel.cvut.cz/ontologies/slovník/agendový/popis-dat/pojem/má-datum-a-čas-modifikace> .
```
ledsoft: It would make sense to use `xsd:dateTime` as range of this property.

src/main/java/cz/cvut/kbss/termit/service/business/ResourceService.java
ledsoft, line 194: Missing javadoc explainin the new parameter.

src/main/java/cz/cvut/kbss/termit/service/document/backup/DocumentBackupManager.java
```diff
@@ -39,11 +43,26 @@
...
         this.storageDirectory = Path.of(config.getFile().getStorage());
+        this.resourceRepositoryService = resourceRepositoryService;
+    }
+
+    /**
+     * Maps {@link ResourceSaveReason} to {@link BackupReason} if the backup should be created.
+     * @param saveReason The save reason to map
+     * @return The mapped value or empty optional if the backup should not be created.
+     */
+    public static Optional<BackupReason> mapSaveReasonToBackup(ResourceSaveReason saveReason) {
```
ledsoft: Why is the mapping not part of one of the enums? For example, constants in `BackupReason` could have a corresponding `ResourceSaveReason` (if applicable).

src/main/java/cz/cvut/kbss/termit/service/document/backup/DocumentScheduledBackupManager.java
```diff
@@ -0,0 +1,59 @@
...
+import cz.cvut.kbss.termit.model.resource.File_;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Service;
+
+import java.util.List;
+
+@Service
+public class DocumentScheduledBackupManager {
```
ledsoft: `ScheduledDocumentBackupManager` sounds better IMO.

src/main/java/cz/cvut/kbss/termit/service/document/backup/DocumentScheduledBackupManager.java
```diff
@@ -0,0 +1,59 @@
...
+        for (File file : toBackup) {
+            try {
+                backupManager.createBackup(file, BackupReason.SCHEDULED);
+            } catch (Exception e) {
+                LOG.error("Failed to create scheduled backup for file <{}>", file.getUri(), e);
+            }
+        }
+    }
+
+    List<File> findFilesToBackup() {
```
ledsoft: I would prefer this kind of method (using SPARQL, calling JOPA) in the `ResourceDao`.

src/main/java/cz/cvut/kbss/termit/service/document/backup/DocumentScheduledBackupManager.java
ledsoft: Could you add a short explanatory Javadoc for the class itself? I try to write javadoc to classes where the use case is not absolutely clear from the name (like DAO, entity service)...

ledsoft: The comment could, for example, mention that this class handles regular backup of files

ledsoft (changes requested):
See the inline comments for change suggestions.

ledsoft (approved):
LGTM

PR #365, opened 2025-11-17
[Feat #702] Extend custom attributes for annotating term relationships

### LINKED ISSUE #702: Extend custom attributes for annotating term relationships
Custom attributes need to be extended for the annotation of term relationships. This extension consists of:

- New domain of the custom attribute - "term relationships" (stored as `rdf:Statement`)
- For such custom attributes, the user can select from a set of properties representing term attributes to which the custom attribute will be applicable. The set of properties will consist of: 
  - basic term attributes, 
  - custom attributes whose domain and range are `skos:Concept`

Haschman:


ledsoft (commented):
What is the motivation of moving domain and range from `CustomAttribute` to `RdfsResources`? kbss-cvut/termit-ui#702 clearly states that only known SKOS-based mapping term properties (such as `skos:related`, `skos:relatedMatch` etc. - these can be enumerated in code) and user defined custom attributes (i.e., not any user-defined properties) should be considered for selection of annotated properties, so I do not see any reason why retrieval of all properties (in `DataDao`) should consider domains and ranges.

Haschman:
> What is the motivation of moving domain and range from `CustomAttribute` to `RdfsResources`? [kbss-cvut/termit-ui#702](https://github.com/kbss-cvut/termit-ui/issues/702) clearly states that only known SKOS-based mapping term properties (such as `skos:related`, `skos:relatedMatch` etc. - these can be enumerated in code) and user defined custom attributes (i.e., not any user-defined properties) should be considered for selection of annotated properties, so I do not see any reason why retrieval of all properties (in `DataDao`) should consider domains and ranges.

It is used for adding additional info about domain and range to properties. This way when we are fetching all properties we can filter those that: "domain and range are skos:Concept". Why hard code it on the front-end when we can fetch them from the back-end?

ledsoft:
> > What is the motivation of moving domain and range from `CustomAttribute` to `RdfsResources`? [kbss-cvut/termit-ui#702](https://github.com/kbss-cvut/termit-ui/issues/702) clearly states that only known SKOS-based mapping term properties (such as `skos:related`, `skos:relatedMatch` etc. - these can be enumerated in code) and user defined custom attributes (i.e., not any user-defined properties) should be considered for selection of annotated properties, so I do not see any reason why retrieval of all properties (in `DataDao`) should consider domains and ranges.
> 
> It is used for adding additional info about domain and range to properties. This way when we are fetching all properties we can filter those that: "domain and range are skos:Concept". Why hard code it on the front-end when we can fetch them from the back-end?

I see a couple of caveats/points to mention:
- The set of SKOS mapping properties is fixed and will not change, so hardcoding it is not such a big deal. What can change are the user-defined custom attributes and these are/should be fetched
- Putting domain and range into RDFS Resource is factually incorrect because a resource does not have these, they are defined for RDF Property
- It adds the necessity of using the UNION for getting domains/ranges of parent properties (I checked the specs and it cannot be inferred unless OWL entailment is used, which we don't have) 

I sort of like your reasoning, but in that case we should IMHO introduce a class representing an RDF Property and add the domain/range to it (there would basically be a inheritance chain `CustomAttribute subClassOf RdfProperty subClassOf RdfsResource`)

ledsoft (approved):
LGTM

PR #366, opened 2025-11-21
[Fix kbss-cvut/termit-ui#702] CustomAttribute update to include annotated relationships

### LINKED ISSUE #702: Extend custom attributes for annotating term relationships

Haschman:
…ated relationships


ledsoft (approved):
LGTM

PR #371, opened 2025-11-29
[Enhancement kbss-cvut/termit-ui#679] Provide access to document backups

### LINKED ISSUE #679: Provide access to document backups
As a user, I want to be able to see backups of a file and be able to return to a selected version of the file. 

- The list should indicate the general type of the backup (text analysis invocation, added occurrence, periodic).
- When a backup is restored, TermIt should also generate corresponding term occurrences into the repository (much like when processing the results of text analysis).
- When the backup is restored, the system should first check whether the current version differs from the last available backup. If so, it should create a new backup before applying the selected one.

This follows kbss-cvut/termit#338.

lukaskabc:
Implements kbss-cvut/termit-ui#679

- Adds endpoints for listing backups, their count, and backup restoration
- Restoring a backup will create a new backup of the  current state if the file was modified since the last backup
- Restoring a backup will drop all term occurrences targeting the file and will generate them again based on the file contents
- Backups that do not use the backup name separator (`~`), which was introduced in 2023, will not be listed and cannot be restored

### DIFF
--- a/ontology/termit-glosář.ttl
+++ b/ontology/termit-glosář.ttl
@@ -698,3 +698,27 @@ termit-pojem:má-datum-a-čas-poslední-zálohy
                 termit:glosář ;
         <http://www.w3.org/2004/02/skos/core#prefLabel>
                 "Has date and time of the last backup"@en , "Má datum a čas poslední zálohy"@cs .
+
+termit-pojem:popis-zálohy-souboru
+        a       <http://www.w3.org/2004/02/skos/core#Concept> ;
+        <http://www.w3.org/2004/02/skos/core#broader>
+                <https://slovník.gov.cz/základní/pojem/typ-objektu> ;
+        <http://www.w3.org/2004/02/skos/core#inScheme>
+                termit:glosář ;
+        <http://www.w3.org/2004/02/skos/core#prefLabel>
+                "File backup description"@en , "Popis zálohy souboru"@cs ;
+        <http://www.w3.org/2004/02/skos/core#scopeNote>
+                "Describes a file backup created at a specific time for a specific reason."@en ,
+                "Popisuje zálohu souboru vytvořenou v určitém čase z určitého důvodu."@cs .
+
+termit-pojem:má-důvod-zálohy
+        a       <http://www.w3.org/2004/02/skos/core#Concept> ;
+        <http://www.w3.org/2004/02/skos/core#broader>
+                <https://slovník.gov.cz/základní/pojem/vlastnost> , <https://slovník.gov.cz/základní/pojem/typ-vlastnosti> ;
+        <http://www.w3.org/2004/02/skos/core#inScheme>
+                termit:glosář ;
+        <http://www.w3.org/2004/02/skos/core#prefLabel>
+                "Has backup reason"@en , "Má důvod zálohy"@cs ;
+        <http://www.w3.org/2004/02/skos/core#scopeNote>
+                "Specifies the reason why a file backup was created."@en ,
+                "Určuje důvod, proč byla záloha souboru vytvořena."@cs .
--- a/ontology/termit-glosář.ttl
+++ b/ontology/termit-glosář.ttl
@@ -698,27 +698,3 @@ termit-pojem:má-datum-a-čas-poslední-zálohy
                 termit:glosář ;
         <http://www.w3.org/2004/02/skos/core#prefLabel>
                 "Has date and time of the last backup"@en , "Má datum a čas poslední zálohy"@cs .
-
-termit-pojem:popis-zálohy-souboru
-        a       <http://www.w3.org/2004/02/skos/core#Concept> ;
-        <http://www.w3.org/2004/02/skos/core#broader>
-                <https://slovník.gov.cz/základní/pojem/typ-objektu> ;
-        <http://www.w3.org/2004/02/skos/core#inScheme>
-                termit:glosář ;
-        <http://www.w3.org/2004/02/skos/core#prefLabel>
-                "File backup description"@en , "Popis zálohy souboru"@cs ;
-        <http://www.w3.org/2004/02/skos/core#scopeNote>
-                "Describes a file backup created at a specific time for a specific reason."@en ,
-                "Popisuje zálohu souboru vytvořenou v určitém čase z určitého důvodu."@cs .
-
-termit-pojem:má-důvod-zálohy
-        a       <http://www.w3.org/2004/02/skos/core#Concept> ;
-        <http://www.w3.org/2004/02/skos/core#broader>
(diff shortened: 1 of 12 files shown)


src/main/java/cz/cvut/kbss/termit/service/document/backup/DocumentBackupManager.java
```diff
@@ -241,12 +241,41 @@ private BackupFile resolveFileVersionAt(File file, Instant at) {
...
+     * Restores the given backup file of the file resource.
+     * Backups the original contents if needed.
+     * <p>
+     * <b>Does not resolve occurrences in the restored file.</b>
+     * @param fileResource the file resource
+     * @param backupFile the backup to restore
+     * @return The physical file replaced with the backup.
+     */
+    @Transactional
+    public java.io.File restoreBackup(File fileResource, BackupFile backupFile) {
```
ledsoft, line 264: Please add some logging.

src/main/java/cz/cvut/kbss/termit/rest/ResourceController.java
```diff
@@ -368,6 +374,83 @@ public List<AbstractChangeRecord> getHistory(
...
+            @RequestParam(name = "pageSize", required = false, defaultValue = "-1") int pageSize
+            ) {
+        final Resource resource = getResource(localName, namespace);
+        Pageable pageable = Constants.DEFAULT_PAGE_SPEC;
+        if (pageSize > -1) {
+            pageable = Pageable.ofSize(pageSize);
+        }
+        pageable = pageable.withPage(page);
+
+        final List<FileBackupDto> dtos = resourceService.getBackupFiles(resource).stream()
```
ledsoft, line 422: To me it makes more sense for the service to return the backup DTOs already sorted from the latest to the oldest. I think the `DocumentBackupManager` should return the backups sorted.

src/main/java/cz/cvut/kbss/termit/rest/dto/FileBackupDto.java
```diff
@@ -0,0 +1,79 @@
...
+import java.time.Instant;
+import java.util.Objects;
+
+/**
+ * Dto used for description of a file backup created at {@link #timestamp}
+ * with a {@link #backupReason}
+ */
+@NonEntity
+@OWLClass(iri = Vocabulary.s_c_popis_zalohy_souboru)
+public class FileBackupDto {
```
ledsoft, line 21: Consider implementing this as a record.

src/main/java/cz/cvut/kbss/termit/service/business/ResourceService.java
```diff
@@ -389,4 +403,42 @@ public List<AbstractChangeRecord> getChanges(Resource asset, ChangeRecordFilterD
...
+     * Restores a backup for the given resource from the given timestamp.
+     * If no backup was created at the exact timestamp, the closest newest backup is restored.
+     *
+     * @param resource the resource for which the backup should be restored; must be a {@link File}
+     * @param backupTimestamp the timestamp of the desired backup to restore
+     */
+    @Transactional
+    @Throttle(value = "{#resource.getUri()}")
+    @PreAuthorize("@resourceAuthorizationService.canModify(#resource)")
+    public void restoreBackup(Resource resource, Instant backupTimestamp) {
```
ledsoft, line 430: Please add some logging.

ledsoft (commented):
See the inline comments. I also asked Michal whether it makes sense to add new concepts representing only a DTO (and thus never stored in the repository) to the TermIt ontology.

src/main/java/cz/cvut/kbss/termit/rest/dto/FileBackupDto.java
lukaskabc: I will wait for a follow-up from Michal. If we remove the annotations, I will convert it to a record.

ledsoft: Oh wait, I don't know if jb4json-ld supports serialization of records. Disregard this, sorry.

ontology/termit-glosář.ttl
```diff
@@ -698,3 +698,27 @@ termit-pojem:má-datum-a-čas-poslední-zálohy
...
+        a       <http://www.w3.org/2004/02/skos/core#Concept> ;
+        <http://www.w3.org/2004/02/skos/core#broader>
+                <https://slovník.gov.cz/základní/pojem/vlastnost> , <https://slovník.gov.cz/základní/pojem/typ-vlastnosti> ;
+        <http://www.w3.org/2004/02/skos/core#inScheme>
+                termit:glosář ;
+        <http://www.w3.org/2004/02/skos/core#prefLabel>
+                "Has backup reason"@en , "Má důvod zálohy"@cs ;
+        <http://www.w3.org/2004/02/skos/core#scopeNote>
+                "Specifies the reason why a file backup was created."@en ,
+                "Určuje důvod, proč byla záloha souboru vytvořena."@cs .
```
ledsoft: After discussion with Michal, we agreed not to extend the ontology with these terms when they are not used to actually store the data. Use the TermIt ontology as a base for the IRIs (see, e.g., `PasswordChangeDto`) and reuse existing vocabularies if possible (e.g., DC Terms description for "popis-zálohy-souboru").

src/main/java/cz/cvut/kbss/termit/service/document/backup/DocumentBackupManager.java
```diff
@@ -241,12 +242,43 @@ private BackupFile resolveFileVersionAt(File file, Instant at) {
...
+        // create backup if the file was modified since the last backup
+        if (fileResource.getModified().isAfter(fileResource.getLastBackup())) {
+            createBackup(fileResource, BackupReason.BACKUP_RESTORE);
+        }
```
ledsoft, line 270: I think it also makes sense to log info whether the current version is backed up or whether it is not necessary to do so.

lukaskabc: The createBackup method will log that a backup is being created with its file name, which also contains the reason (line 87). Should I add another log here anyway?

https://github.com/kbss-cvut/termit/pull/371/files/9bed79ec0d347559a81eeab0edc35953d61271f5#diff-6694f50c2828e5fce60e713b7acc9a0582c310b793a0fffaaa796a03c8dd585cR87

ledsoft: I was a little confused when the backup was not created. Only then I realized that we agreed no to create a backup when it does not differ from previous one. Maybe just log debug that there is no need to create a backup in this case.

ledsoft (approved):
LGTM

PR #372, opened 2025-11-30
[Feat kbss-cvut/termit-ui#705] Extend faceted search by annotated relationships

### LINKED ISSUE #705: Extend faceted search with support for annotated term relationships
Faceted search should be extended with support for searching for terms whose relationships are annotated by the selected term.

For example, consider a relationship between terms `A` and `B` and its annotation by term `C`: `<< A skos:related B >> ex:annotatedBy C`. Now if we search for terms with relationships annotated by `C`, `A` and `B` would be in the results.

Haschman:


ledsoft (approved):
LGTM

PR #381, opened 2026-01-27
[4.4.0] Release

ledsoft:


ledsoft:
gitguardian warning is irrelevant - the Actuator password is now generated randomly on startup if not configured by user.

PR #396, opened 2026-03-01
[Feat kbss-cvut/termit-ui#713] Implement advanced search combining full-text search with faceted filtering

### LINKED ISSUE #713: Extend faceted search with FTS
# Popis
Faceted search should support full-text search as well, so that FTS results can be further filtered by facets.

# Využití
As a user, I want to use FTS but I want to filter the results using facets.

# Usefulness

2. It slows me down (I could run FTS and then go through the results manually, but this can be quite time consuming)
****

Haschman:


src/main/java/cz/cvut/kbss/termit/rest/SearchController.java
```diff
@@ -90,22 +88,32 @@ public List<FullTextSearchResult> fullTextSearchTerms(
...
-                                                                     required = false) Integer pageNo,
-                                                       @Parameter(description = "Search parameters.")
-                                                       @RequestBody Collection<SearchParam> searchParams) {
-        if (searchParams.isEmpty()) {
-            throw new UnsupportedSearchFacetException("Search params must be provided for faceted search.");
-        }
-        return searchService.facetedTermSearch(searchParams, RestUtils.createPageRequest(pageSize, pageNo));
+    @PostMapping(value = "/advanced", produces = {MediaType.APPLICATION_JSON_VALUE, JsonLd.MEDIA_TYPE},
+            consumes = {MediaType.APPLICATION_JSON_VALUE})
+    public List<FullTextSearchResult> advancedSearch(
```
ledsoft, line 107: Are the other two endpoints still relevant at this point? Would it make more sense to just have a single `/search` endpoint?

ledsoft: On a second thought - let's keep the endpoints so that the API is backwards compatible, but the endpoint methods could call the advancedSearch service method and we could get rid of all the underlying fts and facetedSearch method, correct?

ledsoft (commented):
Haven't tested the functionality, but from just reviewing the code, I suggest more simplification (feel free to correct me if my assumptions were wrong).

src/main/java/cz/cvut/kbss/termit/rest/SearchController.java
```diff
@@ -72,7 +71,7 @@ public List<FullTextSearchResult> fullTextSearch(@Parameter(description = "Searc
                                                      @Parameter(description = "Search language. " +
                                                              "Searches in all languages if the field is omitted.")
                                                      @RequestParam(name = "language", required = false) String language) {
-        return searchService.fullTextSearch(searchString, language);
+        return searchService.advancedSearch(searchString, language, Collections.emptyList(), RestUtils.createPageRequest(null, null));
```
ledsoft:
```suggestion
        return searchService.advancedSearch(searchString, language, Collections.emptyList(), Constants.DEFAULT_PAGE_SPEC);
```

ledsoft (commented):
One minor comment.
Also maybe there will need to be some small adjustments based on the [issue](https://github.com/kbss-cvut/termit-ui/pull/758#issuecomment-4004562807) I found in the frontend PR?

ledsoft (approved):
LGTM, nice work!

PR #421, opened 2026-08-26
[Bug] Fix FTS not matching certain queries in Czech language

### LINKED ISSUE #835: Full text search does not return results in Czech
# Description
When running full text search (FTS) with a selected language, TermIt does not return matches even though there exist terms with matching label. This issue seems to be specific for Czech.

For example, on KBSS development instance, search for "lokalita" with Czech as search language. TermIt does not return any results even though there are terms with label "Lokalita" in Czech. Removing Czech as the search language returns a number of matches. Searching for "locality" in English also returns matches.

# Severity
2. It is a serious issue

# Alternative solutions
Search without specifying language

lukaskabc:
kbss-cvut/termit-ui#835

Word `lokalita` is indexed by Lucene after processing (stemming) as `lokalit`  
Wildcard query `term*` instructs Lucene to perform prefix matching on indexed terms. There is no further processing of the queried term. The prefix is being matched as specified by the query.

The query `lokalit*` yields results as expected because the query matches indexed terms with the given prefix.
The query `lokalita*` yields no results because there is simply no indexed term with the given prefix.

This PR updates the Lucene query construction to use the format `(+term) OR (+term*)`.
`+term` allows Lucene to perform its text processing
`+term*` keeps the prefix matching  

When there are multiple terms, the query is constructed as `(+a +b +c) AND ((+term) OR (+term*))`

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/persistence/dao/SearchDao.java
+++ b/src/main/java/cz/cvut/kbss/termit/persistence/dao/SearchDao.java
@@ -95,15 +95,75 @@ public class SearchDao {
             return searchString;
         }
         final String[] split = searchString.trim().split("\\s+");
-        final int lastIdx = split.length - 1;
-        if (split[lastIdx].length() >= WILDCARD_MIN_LENGTH) {
-            split[lastIdx] = split[lastIdx] + LUCENE_WILDCARD;
-        }
+
         // Require every token (AND) to avoid huge OR candidate sets that are expensive to rank
         for (int i = 0; i < split.length; i++) {
             split[i] = "+" + split[i];
         }
+
+        final int lastIdx = split.length - 1;
+        // -1 to compensate prepended '+'
+        if (split[lastIdx].length() - 1 >= WILDCARD_MIN_LENGTH) {
+            return joinLuceneQueryWithWildcard(split, searchString.length() + split.length);
+        }
         return String.join(" ", split);
+    }
+
+    /**
+     * Appends string following format {@code (token) OR (token*)} to the {@code builder}.
+     * <p>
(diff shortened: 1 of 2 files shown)


src/main/java/cz/cvut/kbss/termit/persistence/dao/SearchDao.java
```diff
@@ -95,17 +95,77 @@ private static String buildLuceneQueryString(String searchString) {
...
+     */
+    static String joinLuceneQueryWithWildcard(String[] tokens, int totalLength) {
+        int lastIndex = tokens.length - 1;
+        int lastTokenLength = tokens[lastIndex].length();
+
+        // totalLength = every token is appended
+        // lastTokenLength = the last token is appended once more
+        // "() AND (() OR (*))".length = 18
+        int additionalCharCount = "() OR (*)".length() + (tokens.length - 1) * "() AND ()".length();
+        StringBuilder builder = new StringBuilder(totalLength + lastTokenLength + additionalCharCount);
```
ledsoft: This kind of optimization attempt is IMHO not beneficial at all. You construct `String`s only to get their length so it can be passed to a `StringBuilder` constructor. Just use the length available (`totalLength` and possibly `lastTokenLength` and let the `StringBuilder` enlarge the underlying array as necessary)

ledsoft (changes requested):
Works fine, just please address the micro-optimization attempt comment.

ledsoft (approved):
LGTM

src/main/java/cz/cvut/kbss/termit/persistence/dao/SearchDao.java
lukaskabc: > You construct Strings only to get their length

Since they are simple literals defined at compile time, they should be stored in the [String Constant Pool](https://www.geeksforgeeks.org/java/string-constant-pool-in-java/)

Micro-optimization removed.

PR #424, opened 2026-08-31
[Enhancement] Add includeTerms parameter to the public root terms endpoint

### LINKED ISSUE #822: Selected term should be scrolled into view/vybraný pojem by měl být zobrazen ve stromu pojmů
# Description

When a term is selected, it should be scrolled into view in the tree component. For example, when I open the detail of a term, the term tree selector on the right should display the selected term in the tree (not just as a value) - the term should be scrolled into view, and its ancestors should be expanded.

# Usefulness
3. It is a nice to have feature, I can live without it.
***
# Popis
Vybraný pojem by měl být zobrazen ve stromečkové komponentě. Např. když si otevřu detail pojmu, měl by tento pojem být zobrazen v selektoru pojmů vpravo (a ne jen jako jeho hodnota) - menu by mělo nascrollovat na vybraný pojem a expandovat všechny jeho předky.
****

lukaskabc:
Required by https://github.com/kbss-cvut/termit-ui/pull/850


ledsoft (approved):
LGTM

PR #428, opened 2026-09-02
[Enhancement] Update getAll terms endpoint to support includeTerms when retrieving flat list

### LINKED ISSUE #822: Selected term should be scrolled into view/vybraný pojem by měl být zobrazen ve stromu pojmů

lukaskabc:
kbss-cvut/termit-ui#822

In order to allow the tree select component to scroll the selected term in flat-list mode, the flat list endpoint needs to support loading terms with the `includeTerms` parameter.

Support for the `includeTerms` parameter was only added for flat list representation.

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/rest/readonly/ReadOnlyTermController.java
+++ b/src/main/java/cz/cvut/kbss/termit/rest/readonly/ReadOnlyTermController.java
@@ -19,7 +19,6 @@ package cz.cvut.kbss.termit.rest.readonly;
 
 import cz.cvut.kbss.jsonld.JsonLd;
 import cz.cvut.kbss.termit.dto.TermInfo;
-import cz.cvut.kbss.termit.dto.listing.FlatTermDto;
 import cz.cvut.kbss.termit.dto.listing.TermDto;
 import cz.cvut.kbss.termit.dto.readonly.ReadOnlyTerm;
 import cz.cvut.kbss.termit.model.Vocabulary;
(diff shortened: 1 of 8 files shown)


src/main/java/cz/cvut/kbss/termit/service/business/TermService.java
```diff
@@ -761,4 +758,24 @@ public Term findVersionValidAt(Term asset, Instant at) {
...
+
+    /**
+     * Finds all terms from the given vocabulary.
+     *
+     * @param vocabulary the vocabulary from which terms should be returned
+     * @param includeTerms terms that should always be included in the result, no matter on which page
+     * @param selectionParams term search parameters
+     * @return flattened list of terms
+     */
+    @Transactional
```
ledsoft:
```suggestion
    @Transactional(readOnly = true)
```

src/main/java/cz/cvut/kbss/termit/rest/readonly/ReadOnlyTermController.java
```diff
@@ -105,10 +108,17 @@ public List<?> getTerms(@Parameter(description = TermController.ApiDoc.ID_LOCAL_
             return termService.findAll(searchString, vocabulary, new TermSelectionParams(flat, false, includeImported, includeRelated,
                                                                                          createPageRequest(pageSize,
                                                                                                            pageNo)));
-        } else {
-            return termService.findAll(vocabulary, new TermSelectionParams(flat, false, includeImported, includeRelated,
-                                                                           createPageRequest(pageSize, pageNo)));
         }
+
+        if (flat && includeTerms != null && !includeTerms.isEmpty()) {
```
ledsoft: `includeTerms` should not be `null` if you provided a default value AFAIK

ledsoft (commented):
Two minor comments

src/main/java/cz/cvut/kbss/termit/service/business/TermService.java
```diff
@@ -158,6 +158,7 @@ public TermInfo findRequiredTermInfo(URI id) {
      * @param selectionParams Term selection parameters
      * @return Matching terms
      */
+    @Transactional
```
ledsoft:
```suggestion
    @Transactional(readOnly=true)
```
These methods were not originally transactional because they did not coordinate invocation of multiple services (and the underlying service is transactional). So either remove the `@Transactional` altogether or mark the transaction as readonly.

lukaskabc: can't remove since it now calls a local method that needs the transaction, marked as read-only

ledsoft (approved):
LGTM

PR #431, opened 2026-09-07
[Bug] Evict Jopa cache for imported vocabulary

lukaskabc:
When a vocabulary is imported, the importer may return an incomplete object (e.g., with missing root terms). 
Furthermore, when an existing vocabulary is being replaced, the old managed object could be cached by Jopa.

Changes - on each Vocabulary import:
- clear entity manager
- evict entity manager cache for the vocabulary context
- always return a fresh Vocabulary object from the repository


ledsoft (approved):
LGTM

PR #432, opened 2026-09-11
[Ref] Dont put included terms at the end of the list when they are present on the page being loaded

lukaskabc:
Until now, termit would exclude included terms from page results and would append them at the end of the page, breaking the expected sorting order. 
After this change, termit will not load includeTerms if they are already present in the result.


ledsoft (approved):
LGTM

PR #433, opened 2026-09-14
[Enhancement] Support term force removal

lukaskabc:
kbss-cvut/termit-ui#744

Frontend PR: https://github.com/kbss-cvut/termit-ui/pull/864

- Added endpoint for listing references to specific Term
- Updated endpoint for Term removal, allowing the specification of how the term should be removed

### DIFF
--- a/src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
+++ b/src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
@@ -49,10 +49,15 @@ import cz.cvut.kbss.termit.persistence.snapshot.TermSnapshotLoader;
 import cz.cvut.kbss.termit.service.snapshot.SnapshotProvider;
 import cz.cvut.kbss.termit.util.Configuration;
 import cz.cvut.kbss.termit.util.Utils;
+import org.eclipse.rdf4j.model.Literal;
 import org.eclipse.rdf4j.model.Statement;
 import org.eclipse.rdf4j.model.util.Values;
+import org.eclipse.rdf4j.query.BindingSet;
 import org.eclipse.rdf4j.query.GraphQuery;
 import org.eclipse.rdf4j.query.GraphQueryResult;
+import org.eclipse.rdf4j.query.Operation;
+import org.eclipse.rdf4j.query.TupleQuery;
+import org.eclipse.rdf4j.query.TupleQueryResult;
 import org.eclipse.rdf4j.repository.RepositoryConnection;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.event.EventListener;
@@ -86,16 +91,19 @@ public class TermDao extends BaseAssetDao<Term> implements SnapshotProvider<Term
     private static final URI DC_TERMS_LANGUAGE = URI.create(DC.Terms.LANGUAGE);
 
     /**
-     * Matches triples where the term is an object in a vocabulary graph excluding vocabulary snapshots
+     * Matches triples where the term is an object in a vocabulary graph excluding vocabulary snapshots.
      */
     private static final String REFERENCES_TO_TERM_WHERE_CLAUSE = """
                 WHERE {
                     GRAPH ?context {
-                        ?other ?relation ?term .
-                        ?context a ?vocabulary .
                         FILTER NOT EXISTS {
                             ?context a ?versionOfVocabulary .
                         }
+                        ?other ?relation ?term .
+                        ?context a ?vocabulary .
+#                        FILTER (?relation NOT IN (
+#                            <http://www.w3.org/2004/02/skos/core#hasTopConcept>
+#                        ))
                     }
                 }
                 """;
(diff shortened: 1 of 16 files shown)


src/main/java/cz/cvut/kbss/termit/service/repository/term_removal/TermRemovalParams.java
```diff
@@ -0,0 +1,43 @@
+package cz.cvut.kbss.termit.service.repository.term_removal;
+
+import cz.cvut.kbss.termit.model.Term;
+import cz.cvut.kbss.termit.model.Vocabulary;
+
+import java.util.Objects;
+
+/**
+ * Describes how should be a term removed
```
ledsoft:
```suggestion
 * Describes how a term should be removed
```

src/main/java/cz/cvut/kbss/termit/service/repository/term_removal/TermRemovalParamsApplier.java
```diff
@@ -0,0 +1,16 @@
+package cz.cvut.kbss.termit.service.repository.term_removal;
+
+import cz.cvut.kbss.termit.service.repository.TermRepositoryService;
+
+/**
+ * Applier performing the given operation described by the parameters.
+ */
+public interface TermRemovalParamsApplier {
+    /**
+     * Applies the operation described by the parameters.
```
ledsoft:
```suggestion
     * Removes a term, performing additional operations based on the specified parameters.
```

src/main/java/cz/cvut/kbss/termit/service/repository/term_removal/TermRemovalParamsApplier.java
```diff
@@ -0,0 +1,16 @@
+package cz.cvut.kbss.termit.service.repository.term_removal;
+
+import cz.cvut.kbss.termit.service.repository.TermRepositoryService;
+
+/**
+ * Applier performing the given operation described by the parameters.
```
ledsoft:
```suggestion
 * Removes terms and executes any necessary related operations.
```

src/main/java/cz/cvut/kbss/termit/service/repository/TermRepositoryService.java
```diff
@@ -618,6 +643,58 @@ public Optional<URI> findTermVocabulary(URI termId) {
...
+     * and there must be no references to the term.
+     *
+     * @param instance The instance to remove
+     * @see #remove(TermRemovalParams, Vocabulary) 
+     * @see #forceRemove(Term) 
+     */
+    @Override
+    public void remove(Term instance) {
+        super.remove(instance);
+    }
```
ledsoft, line 696: This is not necessary, when it just calls super, is it?

src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
```diff
@@ -1276,4 +1317,154 @@ public Set<TermInfoWithParents> findWithAllAncestors(Set<URI> termUris) {
...
+    }
+
+    /**
+     * Counts the total amount of references to the specified term
+     *
+     * @param con the repository connection
+     * @param term the term to which references should be counted
+     * @return the total number of statements referencing the term as object
+     */
+    private long countReferences(RepositoryConnection con, AbstractTerm term) {
```
ledsoft, line 1376: Why are you using RDF4J API here? JOPA query API can do the same service (inference can be disabled by applying a query hint). I would like to reduce the reliance on RDF4J API to as little as possible.

src/main/java/cz/cvut/kbss/termit/persistence/dao/BaseDao.java
```diff
@@ -162,4 +166,24 @@ protected Descriptor getDescriptor() {
...
     }
+
+    /**
+     * Unwraps the supplied entity manager to {@link RepositoryConnection} within the current active transaction.
+     *
+     * @param entityManager the entity manager to unwrap
+     * @return open repository connection with active current transaction
+     * @implSpec The connection is managed by the {@link EntityManager} and must not be used with try-auto-close statement.
+     */
+    public static RepositoryConnection unwrapToConnection(EntityManager entityManager) {
```
ledsoft, line 177: IMHO, if you used regular JOPA query API to get the reference count, you would have the connection initialized, so getting the `RepositoryConnection` would require only calling `em.unwrap(RepositoryConnection.class)`, which could be done in-place in TermDao and we could remove this utility function completely. Again reducing the numer of places where RDF4J API leaks into TermIt.

ledsoft (changes requested):
Couple of minor code style comments. I also don't like the underscore in `term_removal` package name. I would prefer a single-word name (such as `removal`, the fact that the classes handle Term removal is clear from their name) or a nested package.

ledsoft:
When I remove a term and its child is supposed to become a root term, it does not become one. This causes it to disappear from the vocabulary, although it still exists in the repository. It is just missing the `voc skos:hasTopConcept term` statement

src/main/java/cz/cvut/kbss/termit/service/repository/TermRepositoryService.java
lukaskabc: I added it for javadoc purposes to reference the other remove methods, since this one is from the base dao and does not allow for removing the term with the removal params

src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
lukaskabc: to reuse the query and parameter binding, otherwise I would need to do it twice for the same `WHERE` query, once for RDF4J and once for JOPA

and `findReferences` uses RDF4J to access the statements

src/main/java/cz/cvut/kbss/termit/service/repository/TermRepositoryService.java
```diff
@@ -188,6 +205,8 @@ private void pruneAndNormalizeTranslations(Term instance) {
     @Override
     protected void postUpdate(@Nonnull Term instance) {
         final Vocabulary vocabulary = vocabularyService.findRequired(instance.getVocabulary());
+        // Jopa should always set at least empty collection
+        Objects.requireNonNull(vocabulary.getRootTerms(), "Vocabulary must have root terms collection");
```
ledsoft, line 209: How is this relevant?

src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
```diff
@@ -77,6 +91,25 @@ public class TermDao extends BaseAssetDao<Term> implements SnapshotProvider<Term
...
+     * {@code skos:hasTopConcept} relations are excluded
+     */
+    private static final String REFERENCES_TO_TERM_WHERE_CLAUSE = """
+                WHERE {
+                    GRAPH ?context {
+                        FILTER NOT EXISTS {
+                            ?context a ?versionOfVocabulary .
+                        }
+                        ?other ?relation ?term .
+                        ?context a ?vocabulary .
```
ledsoft, line 105: Since you are already using `skos:hasTopConcept`, just reference `skos:ConceptScheme` here and get rid of the query binding setting.

src/main/java/cz/cvut/kbss/termit/persistence/dao/TermDao.java
ledsoft: Taking into account the suggestion of inlining the vocabulary type in the WHERE condition string, I stand by comment. It would become only two bindings and would allow us to remove the `unwrapToConnection` method from the `BaseDao`, signficantly simplify it and reduce the number of places where RDF4J API is used.

ledsoft (changes requested):
Additional comments after re-review.

ledsoft (approved):
LGTM

PR #435, opened 2026-09-18
[Upd] Update JOPA to 2.12.3

lukaskabc:


lukaskabc:
@ledsoft There is an assertion error from Jopa at [AbstractUnitOfWork:576](https://github.com/kbss-cvut/jopa/blob/76b5dd3be35d84d06ff2d21a56d533dfa5fbf1e1/jopa-impl/src/main/java/cz/cvut/kbss/jopa/sessions/AbstractUnitOfWork.java#L576)

ledsoft (approved):

PR #437, opened 2026-09-23
[Enhancement] Allow reverting asset change

lukaskabc:
Implements https://github.com/kbss-cvut/termit-ui/issues/790

Frontend PR: https://github.com/kbss-cvut/termit-ui/pull/876

# Rollback restrictions:
- Term creation/deletion cannot be rolled back
- Rollbacks are possible only for update change records (modification to asset attributes)
- Rollback is possible only when:
  - The original value was **literal** (number, string...)
  - The original value was a reference to an asset (Term, Vocabulary), and the asset **still exists**.  
 For example, when the original value referenced term that was already deleted, the rollback wont be possible, because the reference would be invalid. Whether the referenced resource is (or was) an asset is being resolved using the metamodel and custom attributes (with Term range).
  - The original value was a resource (URI), and a custom attribute with `Resource (IRI/URI/URL)` range exists.  
    This means the URI is a general resource, and its existence will not be validated.

## About additional properties:
- Additional properties added through termit are currently saved as strings, so their changes can be rolled back
- If some additional property is added in other means (custom attribute is removed, via import...) and its value is saved as IRI, the change won't be possible to rollback, because there is no indication whether it stores a valid reference or not.

Applying rollback creates a new change record and does not change previous change records.

# Changes
This PR adds:
- New change type in the Termit ontology `termit:reversible-change` representing a reversible update of an entity.
- Enhanced `/history` endpoints of `Vocabulary` and `Term` adding `termit:reversible-change` type to change records that can be rolled back by the current user.
- Two new endpoints for rolling back an update change record (again one for Vocabulary, and one for Term).
- `ChangeRollbackService` for performing rollbacks
- `RollbackValidator` validates whether the rollback of the given update change record would be possible without breaking the rules above


lukaskabc:
The PR is put as a draft until https://github.com/kbss-cvut/jopa/issues/483 is resolved, causing the rollback of a MultilingualString property possibly discarding all translations except one.

